PR #1, opened 2019-09-10
Feature/node agent as config

27149chen:
Response to this issue: https://github.com/operator-framework/operator-lifecycle-manager/issues/1022
1. Add a new crd called Config
2. Enable the node agent daemonSet when config.spec.defineHost is set to true.
3. Now the operator only have one deployment and its rbac roles.


oriyarde:

bitvijays:
@27149chen Hope you are doing well. We came across https://github.com/operator-framework/operator-lifecycle-manager/issues/1022 and wanted to ask, was your requirement to install the node-agent automatically (similar to daemonset) at every node using the operator?

As previously, when you would be apply using operator, it would only deploy on one node?

PR #10, opened 2019-10-16
Feature/csi 548 csi driver linter fix

27149chen:
1. add required labels and annotations
2. add node affinity
3. add liveness probe for all containers
4. add cpu/mermory requests and limits
5. set default security context (drop all, add nothing)

### DIFF
--- a/pkg/controller/ibmblockcsi/syncer/csi_controller.go
+++ b/pkg/controller/ibmblockcsi/syncer/csi_controller.go
@@ -20,6 +20,7 @@ import (
 	"github.com/imdario/mergo"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -57,6 +58,7 @@ func NewCSIControllerSyncer(c client.Client, scheme *runtime.Scheme, driver *ibm
 			Name:        config.GetNameForResource(config.CSIController, driver.Name),
 			Namespace:   driver.Namespace,
 			Annotations: driver.GetAnnotations(),
+			Labels:      driver.GetLabels(),
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/ibmblockcsi/syncer/csi_controller.go
```diff
@@ -157,15 +155,64 @@ func (s *csiControllerSyncer) ensureContainersSpec() []corev1.Container {
...
+}
+
+func ensureNodeAffinity() *corev1.NodeAffinity {
+	return &corev1.NodeAffinity{
+		RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+			NodeSelectorTerms: []corev1.NodeSelectorTerm{
+				{
+					MatchExpressions: []corev1.NodeSelectorRequirement{
+						{
+							Key:      "beta.kubernetes.io/arch",
```
shay-berman: please fix the key to kubenetes.io/arch (with no beta prefix - i think its deprecated also from k8s 1.14)

pkg/controller/ibmblockcsi/syncer/csi_controller.go
```diff
@@ -157,15 +155,64 @@ func (s *csiControllerSyncer) ensureContainersSpec() []corev1.Container {
 	}
 }
 
+func ensureDefaultResources() corev1.ResourceRequirements {
+	return ensureResources("50m", "100m", "50Mi", "100Mi")
```
shay-berman: it looks like you put these limits\requests to all the containers of the node pod and controller pod.
Just wondering how you decide about the size here? is there any guideline that state these are the requested\limit for all the containers?  or it just a guess?

Another question, is it more standard to allow the user to set these params as part of the operator parameters? assuming the customer has issue with cpu and he need more, he will not be able to do anything.
do u think its good enough to keep it as this, or at least give the abiliy to configure it if user wants(lets say its a cluster with 200 nodes and with 1000 PVCs maybe 500m cpu is not enough its really loaded)?  thoughts?

pkg/controller/ibmblockcsi/syncer/csi_node.go
```diff
@@ -287,3 +297,17 @@ func ensureHostPathVolumeSource(path, pathType string) corev1.VolumeSource {
 		},
 	}
 }
+
+func fillSecurityContextCapabilities(sc *corev1.SecurityContext, add ...string) {
+	sc.Capabilities = &corev1.Capabilities{
+		Drop: []corev1.Capability{"ALL"},
```
shay-berman: only if we will have time before GA, we can test which specific caps needed for node, but for now ALL is ok.  (if u can please open jira ticket for that so we will not forget it)

pkg/internal/ibmblockcsi/ibmblockcsi.go
```diff
@@ -42,52 +42,65 @@ func (c *IBMBlockCSI) Unwrap() *csiv1.IBMBlockCSI {
...
 }
 
-// GetAnnotations returns all the annotations to be set on all resources
-func (c *IBMBlockCSI) GetAnnotations() labels.Set {
+// GetLabels returns all the labels to be set on all resources
+func (c *IBMBlockCSI) GetLabels() labels.Set {
 	labels := labels.Set{
 		"app.kubernetes.io/name":       config.ProductName,
 		"app.kubernetes.io/instance":   c.Name,
-		"app.kubernetes.io/version":    csiversion.Version,
```
shay-berman: r u sure?   i think its better to keep it as well, since its vanila k8s recommendation.

PR #12, opened 2019-10-17
Fix/arch key + disable serviceMonitoring and clusterDriverRegisterer

shay-berman:
**1. README update**
1.1. put master branch in the curl operations - make it ready for GA.
1.2. remove IBMBlockCSI yaml from the readme, and instead asking the user to create it, we just move it as a file that the user curl and download from github. the same as the operator yaml.
1.3. Use -l csi to get the pods

**2. fix kubernetes.io/arch**  
instead of beta.kubernetes.io/arch which is deprecated.

**3. Disable the ServiceMonitor**
Disable the creation of ServiceMonitoring since we don't have time to test it for v1. Created epic in jira for later version

**4. Upgrade csi-node-driver-registrar:v1.2.0** 
Upgrade to csi-node-driver-registrar:v1.1.0  -> v1.2.0 which also support k8s 1.13+. https://kubernetes-csi.github.io/docs/node-driver-registrar.html

**5. remove cluster-driver-registrar (CSI-480)**
cluster-driver-registrar is officially deprecated -> https://kubernetes-csi.github.io/docs/cluster-driver-registrar.html
So remove it from the operator in all means - as side car container and with any role\binding.

**6. Improve makefile**
Makefile now has default target that compile the project + list target to see all the targets.

**7. missing csi label on operator pod**


cmd/manager/main.go
```diff
@@ -144,6 +144,8 @@ func main() {
 		log.Info("Could not create metrics Service", "error", err.Error())
 	}
 
+	/*
```
27149chen: You should disable serveCRMetrics and CreateMetricsService either

shay-berman: you right
Done, please check again

shay-berman:
@27149chen  - please review again and approve

Makefile
```diff
@@ -14,12 +14,23 @@
 # limitations under the License.
 #
 
-all: test
+#all: test
+
+.PHONY: ibm-block-csi-operator
+ibm-block-csi-operator:
+	mkdir -p bin
```
27149chen: we don't have a path called bin

Makefile
```diff
@@ -14,12 +14,23 @@
...
+	CGO_ENABLED=1 GOOS=linux go build     -o build/_output/bin/ibm-block-csi-operator     -gcflags all=-trimpath=${GOPATH} -asmflags all=-trimpath=${GOPATH} -mod=vendor cmd/node/main.go
```
27149chen: cmd/node/main.go -> cmd/manager/main.go

Makefile
```diff
@@ -14,12 +14,23 @@
...
 
 .PHONY: test
-test: update
```
27149chen: why removing "update"?

Makefile
```diff
@@ -14,12 +14,23 @@
...
+test:
 	ginkgo -r -skipPackage pkg/controller
 
 .PHONY: update
 update:
 	hack/update-all.sh
+
+
+.PHONY: list
+list:
```
27149chen: what's "list" for?

go.mod
```diff
@@ -6,15 +6,18 @@ require (
...
 	github.com/pkg/errors v0.8.1
 	github.com/presslabs/controller-util v0.1.14-0.20190818132821-fc080f9430fa
 	github.com/spf13/pflag v1.0.3
+	golang.org/x/net v0.0.0-20191014212845-da9a3fd4c582 // indirect
+	golang.org/x/sys v0.0.0-20191018095205-727590c5006e // indirect
 	google.golang.org/grpc v1.19.1
+	gopkg.in/yaml.v2 v2.2.4 // indirect
 	k8s.io/api v0.0.0-20190918155943-95b840bb6a1f
 	k8s.io/apimachinery v0.0.0-20190913080033-27d36303b655
 	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
```
27149chen: why go.mod is changed?

shay-berman: maybe because my laptop go version is 1.13.3, and the ginkgo version is 1.10.2.
Do you want me to return it back to the old ginkgo\mega\protobuf versions?  or u want to keep it with the new version? (up to you, you know the dependency process - so what ever you decide is ok by me)

27149chen: did you run go mod tidy or something similiar? I don't think the ginkgo/protobuf version you install should have impact on the go.mod, maybe because of the go version, I will install a go 1.13.3 (I use 1.13.1 now), and see if it is reproducible.
Anyway, please revert this change because it is not related to this PR. If you want to change the go.mod, please create another PR.

Makefile
shay-berman: fixed.

go.mod
shay-berman: We need to build and run testing inside container, to avoid such issues.
anyway for now i revert changes in go.mod\sum
please check again

shay-berman:
@27149chen  please review again - i fixed your comment and also resolved conflict from changes you did in develop, so please check it again.

shay-berman:
Great - merging to develop

PR #14, opened 2019-10-21
Create automatically the ibm CSIDriver CR when installing the driver.

27149chen:
create the CSIDriver object when an IBMBlockCSI is created.


pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -198,6 +199,7 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
 
 	// create the resources which never change if not exist
 	for _, rec := range []reconciler{
+		r.reconcileCSIDriver,
```
shay-berman: The CSIdriver should be created during the creation of the CR and not during the creation of the operator.
just double check with you @27149chen  - at this point in the code it related to the creation of the 
CR right?  if so the its LGTM

PR #15, opened 2019-10-22
remove HostNetwork:true and DNSpolicy on csi node

shay-berman:
Openshift certification indicates that pod.spec.hostNetwork:true on the csi node is not recommended.
I tested it without hostNetwor:true (which means its by default false) and it works well (tested it via the original yamls in the driver PR -> https://github.com/IBM/ibm-block-csi-driver/pull/94).

Since hostNetwork dropped then we can also drop the DNSpolicy setting as well.

### DIFF
--- a/pkg/controller/config/syncer/node_agent.go
+++ b/pkg/controller/config/syncer/node_agent.go
@@ -92,6 +92,8 @@ func (s *nodeAgentSyncer) ensurePodSpec() corev1.PodSpec {
 	return corev1.PodSpec{
 		Containers:  s.ensureContainersSpec(),
 		Volumes:     s.ensureVolumes(),
+		HostNetwork: true,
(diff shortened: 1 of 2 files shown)


pkg/controller/config/syncer/node_agent.go
```diff
@@ -92,8 +92,6 @@ func (s *nodeAgentSyncer) ensurePodSpec() corev1.PodSpec {
 	return corev1.PodSpec{
 		Containers:  s.ensureContainersSpec(),
 		Volumes:     s.ensureVolumes(),
-		HostNetwork: true,
-		DNSPolicy:   "ClusterFirstWithHostNet", // To have DNS options set along with hostNetwork, you have to specify DNS policy explicitly to 'ClusterFirstWithHostNet'
```
27149chen: Please don't touch the node_agent in this PR, it is different. We may need the hostNetwork here. Anyway, we can talk it when it is enabled.

shay-berman:
@27149chen no touch on the node_agent per your request. waiting for your LGTM

PR #21, opened 2019-10-29
set more meaningful resources requests and limits

27149chen:
according to the result in https://jira.xiv.ibm.com/browse/CSI-643

### DIFF
--- a/pkg/controller/ibmblockcsi/syncer/csi_controller.go
+++ b/pkg/controller/ibmblockcsi/syncer/csi_controller.go
@@ -34,11 +34,11 @@ import (
 )
 
 const (
-	socketVolumeName                       = "socket-dir"
-	controllerContainerName                = "ibm-block-csi-controller"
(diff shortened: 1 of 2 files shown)


pkg/controller/ibmblockcsi/syncer/csi_controller.go
```diff
@@ -148,7 +150,7 @@ func (s *csiControllerSyncer) ensureContainersSpec() []corev1.Container {
 }
 
 func ensureDefaultResources() corev1.ResourceRequirements {
-	return ensureResources("50m", "100m", "50Mi", "100Mi")
+	return ensureResources("20m", "40m", "20Mi", "40Mi")
```
shay-berman: r u sure only 40Mi limit?

27149chen: yes, from CSI-643, all the sidecars only need no more than 23m and 20Mi for 3 pods and 300 volumes

PR #23, opened 2019-10-30
reduce rbac rules

27149chen:


deploy/role.yaml
```diff
@@ -11,14 +11,47 @@ rules:
...
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - create
+  - delete
+- apiGroups:
+  - ""
+  resources:
+  - secrets
```
shay-berman: why operator need to get secrets?

deploy/role.yaml
```diff
@@ -11,14 +11,47 @@ rules:
...
   - pods
-  - services
-  - services/finalizers
-  - endpoints
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
```
shay-berman: why need to create configmap?

deploy/role.yaml
```diff
@@ -11,14 +11,47 @@ rules:
...
+  - ""
+  resources:
   - persistentvolumes
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - create
+  - delete
```
shay-berman: why operator need to delete\create PV? its the driver that need that.
what do i miss?

27149chen: Please see my comment under the ticket first, as I said:

>  in order to create all the clusterroles of the driver, the operator's clusterrole needs all the permissions inside these clusterroles, please make sure the rules in the clusterroles of the driver is correct

deploy/role.yaml
27149chen: operator will use it to create a lock for leader selection.

deploy/role.yaml
27149chen: because ibm-block-csi-external-provisioner-role need it

deploy/role.yaml
shay-berman: sure
sorry

deploy/role.yaml
shay-berman: so why it should be here?
this is the operator deployment and not the csi-controller.   do i miss something?

27149chen: as mentioned below, the operator role needs all the rules of the driver, including the csi-external-provisioner

27149chen: 1. Yes, I'm sure. It is easy to understand the reason: the role is to limit the permission of the operator, if I can create a new role in the operator and give more permissions, then there is a big security issue - in fact I can do anything!
4. Good question. I thought about it before. But the operator role has more permissions than the driver, if we use the operator role for the driver, the driver will have too big power.

shay-berman: thanks for the answers
its clear now
Base on your detail it LGTM

PR #26, opened 2019-11-04
expose sidecar images

27149chen:
expose all the sidecar images to IBMBlockCSI crd so that user can change the sidecar images


deploy/crds/csi.ibm.com_ibmblockcsis_crd.yaml
```diff
@@ -60,9 +60,28 @@ spec:
               - repository
               - tag
               type: object
+            sidecars:
```
shay-berman: why the node side cars does not listed inside the "node" tag?

does it make sense to have "sidecars" tag for all the isde cars?

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
```diff
@@ -3,10 +3,27 @@ kind: IBMBlockCSI
...
+    app.kubernetes.io/instance: ibm-block-csi-operator
+    app.kubernetes.io/managed-by: ibm-block-csi-operator
 spec:
   controller:
     repository: ibmcom/ibm-block-csi-driver-controller
     tag: "1.0.0"
   node:
     repository: ibmcom/ibm-block-csi-driver-node
     tag: "1.0.0"
+  sidecars:
```
shay-berman: why not
controller:
    csi-controller:
         repo...
         tag...
     sidecar1
     sidecar2
    ....
    csi-node:
        csi-node:
           ...
           ...
        sidecar1(with the full name of cause)


What do you think?  so there is logic between the side cars

27149chen: for some reasons:
1. Flatter and simpler structure
2. For compatibility reasons, we may add new componts except controller and node, add or remove some of the sidecars, even redesign the controller and node components.
3. Less input, like the liveness probe, we don't need to specify it twice.

shay-berman:
please check some best practice before apply.

maybe review this links
https://github.com/openshift/csi-operator?files=1

https://github.com/nutanix/helm/blob/master/charts/nutanix-csi-storage/values.yaml

we need to make the structure good enough.
so it will be clear what images are and also think about the future to make it with
tolaration, resource limits, pullpolicy and othe aspects that we will have to add in the future.

for example pullpolicy sound important also for v1.


another question, what happend if in the future we will have to update the structure of the CR,   what is the implication?  does it mean customer need to uninstall the rc and install it or othe upgrade path can be found.

shay-berman:
please check some best practice before apply.

1. maybe review this links
https://github.com/openshift/csi-operator?files=1

https://github.com/nutanix/helm/blob/master/charts/nutanix-csi-storage/values.yaml

2  we need to make the structure good enough.
so it will be clear what images are and also think about the future to make it with
tolaration, resource limits, pullpolicy and othe aspects that we will have to add in the future.

3. for example pullpolicy sound important also for v1.

4. what about secret image pull? is it important to expose?


5. another question, what happend if in the future we will have to update the structure of the CR,   what is the implication?  does it mean customer need to uninstall the rc and install it or othe upgrade path can be found.

27149chen:
> please check some best practice before apply.
> 
>     1. maybe review this links
>        https://github.com/openshift/csi-operator?files=1
> 
> 
> https://github.com/nutanix/helm/blob/master/charts/nutanix-csi-storage/values.yaml
> 
> 2 we need to make the structure good enough.
> so it will be clear what images are and also think about the future to make it with
> tolaration, resource limits, pullpolicy and othe aspects that we will have to add in the future.
> 
>     1. for example pullpolicy sound important also for v1.
> 
>     2. what about secret image pull? is it important to expose?
> 
>     3. another question, what happend if in the future we will have to update the structure of the CR,   what is the implication?  does it mean customer need to uninstall the rc and install it or othe upgrade path can be found.
> 
>     4. for example v1.1 we need to add snappshotter, then how the upgrade will look like?  delete the crd and operator and the install the new one or better upgrade path?
> 
>     5. maybe
>        spec:
>        csi-controller-template:
>        ibm-block-csi-driver-controller:
>        repository: ibmcom/ibm-block-csi-driver-controller
>        tag: "1.0.0"
>        pullpolicy: yyy
>        csi-attacher(the orig name):
>        repo....
>        tag....
>        pullpolicy...
>        resource limits....(in the future)
>        csiprovisioner  ..
>        snap in the future...
>        tolaration for the whole template...
> 
> 
> and the same for node template.
> so we donot hide anything and its standard structure.

1. the first link is out of date, no changes for one year, the second link is a helm chart value, not help much
2. right, it is important to keep the structure easy to extend, we can extand the tolaration, resources and other aspects in the current structure.
3. Not sure for v1
4. not sure for v1
5. Good question! there are two situations here:
a. version is not changed. if new fields are added and have a default value: we will set the default value in the operator so that old CRs are updated and compatible after new operator is up, if old fields are remove or renamed: it is not recomanded, better way is marking it as duplicated and remove it when version is changed, and for rename, mark the old one as duplicated and add a new one.
b. version is changed. If two CRDs are not comptible, we need to update version. CRD supoort multiple versioning, we may need a webhook the covert CRs between two versions, see details in https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/
6. since the sidecars is a list, the crd keeps no change, we only need to upgrade the operator
7. I don't think it is a good idea to use the sidecar name as a key, because it is changing, we may add new (like snapshoter), or remove old (like controller register). If we use a specific sidecar as a key, the crd will change again and again.

27149chen:
> > > please check some best practice before apply.
> > > ```
> > > 1. maybe review this links
> > >    https://github.com/openshift/csi-operator?files=1
> > > ```
> > > 
> > > 
> > > https://github.com/nutanix/helm/blob/master/charts/nutanix-csi-storage/values.yaml
> > > 2 we need to make the structure good enough.
> > > so it will be clear what images are and also think about the future to make it with
> > > tolaration, resource limits, pullpolicy and othe aspects that we will have to add in the future.
> > > ```
> > > 1. for example pullpolicy sound important also for v1.
> > > 
> > > 2. what about secret image pull? is it important to expose?
> > > 
> > > 3. another question, what happend if in the future we will have to update the structure of the CR,   what is the implication?  does it mean customer need to uninstall the rc and install it or othe upgrade path can be found.
> > > 
> > > 4. for example v1.1 we need to add snappshotter, then how the upgrade will look like?  delete the crd and operator and the install the new one or better upgrade path?
> > > 
> > > 5. maybe
> > >    spec:
> > >    csi-controller-template:
> > >    ibm-block-csi-driver-controller:
> > >    repository: ibmcom/ibm-block-csi-driver-controller
> > >    tag: "1.0.0"
> > >    pullpolicy: yyy
> > >    csi-attacher(the orig name):
> > >    repo....
> > >    tag....
> > >    pullpolicy...
> > >    resource limits....(in the future)
> > >    csiprovisioner  ..
> > >    snap in the future...
> > >    tolaration for the whole template...
> > > ```
> > > 
> > > 
> > > and the same for node template.
> > > so we donot hide anything and its standard structure.
> > 
> > 
> > 
> > 1. the first link is out of date, no changes for one year, the second link is a helm chart value, not help much
> > 2. right, it is important to keep the structure easy to extend, we can extand the tolaration, resources and other aspects in the current structure.
> > 3. Not sure for v1
> > 4. not sure for v1
> > 5. Good question! there are two situations here:
> >    a. version is not changed. if new fields are added and have a default value: we will set the default value in the operator so that old CRs are updated and compatible after new operator is up, if old fields are remove or renamed: it is not recomanded, better way is marking it as duplicated and remove it when version is changed, and for rename, mark the old one as duplicated and add a new one.
> >    b. version is changed. If two CRDs are not comptible, we need to update version. CRD supoort multiple versioning, we may need a webhook the covert CRs between two versions, see details in https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/
> > 6. since the sidecars is a list, the crd keeps no change, we only need to upgrade the operator
> > 7. I don't think it is a good idea to use the sidecar name as a key, because it is changing, we may add new (like snapshoter), or remove old (like controller register). If we use a specific sidecar as a key, the crd will change again and again.
> 
>     1. how would we will add tolaration in the struction you suggested?    its in the template\pod level and not container level. So how it will look like?
>        Having templateSpecController   and templateSpecNode  as keys  will simplify the adding tolaration.
>        How would you say that tolaration relevant for linessprob of the node and not the liveness probe of the controller if its 1 list?
> 
>     2. Is there any best practice here for setting container image \ tolaration \ resources in operator? (i think it should be the same practice like helm chart)
> 
>     3. i must say i don't have clear answer here, because its really end of the release and we have no much time to deep dive into this. But the side cars must be inside the CR yaml.  and we should consider carfully what is the structure because changing it will cause us troubles in the future.

1. we know that Spectrum Virtualize has a public Cloud version, and supoorts several public cloud platforms, and my question is: do we have a plan to support kubernetes or OpenShift? Another question is: OpenShift use ceph or gluster as its default backend storage, is it possible to replace them with IBM product like Spectrum Virtualize?


1. we can add tolaration under controller and/or node section, they are not just a container spec. It is a user interface, user don't need to know the inside structure, just input there requirement here. For example, if they want the node pod to be installed on master nodes, we will expose something like: deployOnMasters: yes/no
tolaration is per pod, not container. I don't think there will be any different between the liveness container of controller and node, If they do, just give them different names in the sidecars list, like [controller-liveness, node-liveness]
2. go through all the awesome operators in https://github.com/operator-framework/awesome-operators, for the operator-sdk generated operators, I find only 4 CRs that specify resources/tolarations, so I don't think there is a best practice. Inside these 4 CRs, they use them in this way: If there is more than one component, specify them for each component; If there is only one, specify them under .spec
https://github.com/jenkinsci/kubernetes-operator/blob/master/deploy/crds/jenkins_v1alpha2_jenkins_cr.yaml
https://github.com/percona/percona-xtradb-cluster-operator/blob/master/deploy/cr.yaml
https://github.com/lukasz-antoniak/neo4j-operator/blob/master/deploy/crds/database_v1alpha1_neo4jcluster_cr.yaml
https://github.com/storageos/cluster-operator/blob/master/deploy/crds/storageos_v1_storageoscluster_cr.yaml
3. agreed we should be careful about the structure

matt-levan:
For livenessProbe I think the same version should be used by both for v1.  We don't want different versions to accidentally be used at this time.
Global imagePullSecret and pullPolicy would be a good features for v1.
nodeSelector option should be exposed for node:.  There may be cases were not all worker nodes are storage attached.

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
shay-berman: ?

shay-berman:
> > > > please check some best practice before apply.
> > > > ```
> > > > 1. maybe review this links
> > > >    https://github.com/openshift/csi-operator?files=1
> > > > ```
> > > > 
> > > > 
> > > > https://github.com/nutanix/helm/blob/master/charts/nutanix-csi-storage/values.yaml
> > > > 2 we need to make the structure good enough.
> > > > so it will be clear what images are and also think about the future to make it with
> > > > tolaration, resource limits, pullpolicy and othe aspects that we will have to add in the future.
> > > > ```
> > > > 1. for example pullpolicy sound important also for v1.
> > > > 
> > > > 2. what about secret image pull? is it important to expose?
> > > > 
> > > > 3. another question, what happend if in the future we will have to update the structure of the CR,   what is the implication?  does it mean customer need to uninstall the rc and install it or othe upgrade path can be found.
> > > > 
> > > > 4. for example v1.1 we need to add snappshotter, then how the upgrade will look like?  delete the crd and operator and the install the new one or better upgrade path?
> > > > 
> > > > 5. maybe
> > > >    spec:
> > > >    csi-controller-template:
> > > >    ibm-block-csi-driver-controller:
> > > >    repository: ibmcom/ibm-block-csi-driver-controller
> > > >    tag: "1.0.0"
> > > >    pullpolicy: yyy
> > > >    csi-attacher(the orig name):
> > > >    repo....
> > > >    tag....
> > > >    pullpolicy...
> > > >    resource limits....(in the future)
> > > >    csiprovisioner  ..
> > > >    snap in the future...
> > > >    tolaration for the whole template...
> > > > ```
> > > > 
> > > > 
> > > > and the same for node template.
> > > > so we donot hide anything and its standard structure.
> > > 
> > > 
> > > 
> > > 1. the first link is out of date, no changes for one year, the second link is a helm chart value, not help much
> > > 2. right, it is important to keep the structure easy to extend, we can extand the tolaration, resources and other aspects in the current structure.
> > > 3. Not sure for v1
> > > 4. not sure for v1
> > > 5. Good question! there are two situations here:
> > >    a. version is not changed. if new fields are added and have a default value: we will set the default value in the operator so that old CRs are updated and compatible after new operator is up, if old fields are remove or renamed: it is not recomanded, better way is marking it as duplicated and remove it when version is changed, and for rename, mark the old one as duplicated and add a new one.
> > >    b. version is changed. If two CRDs are not comptible, we need to update version. CRD supoort multiple versioning, we may need a webhook the covert CRs between two versions, see details in https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/
> > > 6. since the sidecars is a list, the crd keeps no change, we only need to upgrade the operator
> > > 7. I don't think it is a good idea to use the sidecar name as a key, because it is changing, we may add new (like snapshoter), or remove old (like controller register). If we use a specific sidecar as a key, the crd will change again and again.
> > 
> > 
> > ```
> > 1. how would we will add tolaration in the struction you suggested?    its in the template\pod level and not container level. So how it will look like?
> >    Having templateSpecController   and templateSpecNode  as keys  will simplify the adding tolaration.
> >    How would you say that tolaration relevant for linessprob of the node and not the liveness probe of the controller if its 1 list?
> > 
> > 2. Is there any best practice here for setting container image \ tolaration \ resources in operator? (i think it should be the same practice like helm chart)
> > 
> > 3. i must say i don't have clear answer here, because its really end of the release and we have no much time to deep dive into this. But the side cars must be inside the CR yaml.  and we should consider carfully what is the structure because changing it will cause us troubles in the future.
> > ```
> 
> 1. we can add tolaration under controller and/or node section, they are not just a container spec. It is a user interface, user don't need to know the inside structure, just input there requirement here. For example, if they want the node pod to be installed on master nodes, we will expose something like: deployOnMasters: yes/no
>    tolaration is per pod, not container. I don't think there will be any different between the liveness container of controller and node, If they do, just give them different names in the sidecars list, like [controller-liveness, node-liveness]
> 2. go through all the awesome operators in https://github.com/operator-framework/awesome-operators, for the operator-sdk generated operators, I find only 4 CRs that specify resources/tolarations, so I don't think there is a best practice. Inside these 4 CRs, they use them in this way: If there is more than one component, specify them for each component; If there is only one, specify them under .spec
>    https://github.com/jenkinsci/kubernetes-operator/blob/master/deploy/crds/jenkins_v1alpha2_jenkins_cr.yaml
>    https://github.com/percona/percona-xtradb-cluster-operator/blob/master/deploy/cr.yaml
>    https://github.com/lukasz-antoniak/neo4j-operator/blob/master/deploy/crds/database_v1alpha1_neo4jcluster_cr.yaml
>    https://github.com/storageos/cluster-operator/blob/master/deploy/crds/storageos_v1_storageoscluster_cr.yaml
> 3. agreed we should be careful about the structure


========================

1. ok keep it sidecars structure as is
    the tolaration we will put in the node section or controller section.
    resources can be on each container section and pull policy and pullsecret.
    are you sure we can release v1 without pullsecret section?   maybe its mandatory on openshift?
    
2. good review
    lets keep it as you suggested.

shay-berman:
> For livenessProbe I think the same version should be used by both for v1. We don't want different versions to accidentally be used at this time.
> Global imagePullSecret and pullPolicy would be a good features for v1.
> nodeSelector option should be exposed for node:. There may be cases were not all worker nodes are storage attached.

Agree with Matt
1. imagePullSecret and pullPolicy - i think its mandatory. (does it for openshift?) - can you add it tomorrow?
2. nodeSelector - I think more tolaration then nodeselector. - @27149chen  how much time take to add it? can we do both tomorrow morning and pass staging?

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
```diff
@@ -3,10 +3,56 @@ kind: IBMBlockCSI
...
+    repository: quay.io/k8scsi/csi-attacher
+    tag: "v1.2.1"
+    imagePullPolicy: IfNotPresent
+  - name: livenessprobe
+    repository: quay.io/k8scsi/livenessprobe
+    tag: "v1.1.0"
+    imagePullPolicy: IfNotPresent
+
+  imagePullSecrets:
+  - "secret1"
```
shay-berman: why you put secret1 ????
it should be in comment out  these 2 lines. just if needed then user will add it.
if u keep secret1 then user will have to delete it.

thoughs?

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
```diff
@@ -3,10 +3,56 @@ kind: IBMBlockCSI
...
+                - key: kubernetes.io/arch
+                  operator: In
+                  values:
+                    - amd64
+
   node:
     repository: ibmcom/ibm-block-csi-driver-node
     tag: "1.0.0"
+    imagePullPolicy: IfNotPresent
+    affinity:
```
shay-berman: maybe put comment and say that affinity will impact also the following side car container of the csi node. So it will be clear that other images below also impacted by this setting.,

agree?

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
27149chen: ok, I will comment it out

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
27149chen: affinity is per pod, not per container, I don't think we should say that.

matt-levan: Maybe we should add a comment to the controller and node like.
# Controller pod is created as a statefulSet with the ibm-block-csi-driver-controller and these sidecars(...)
# Node pod is created as daemonSet with the ibm-block-csi-driver-node and these sidecars (...)

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
matt-levan: I do not see where the default serviceaccount (used by the node pod) would get updated with the imagePullSecret.  Is there a reason a serviceAccount isn't used by node?

27149chen: I create a new sa for node, I think it is not a good practice to use the default sa, because it is uncontrollable. Also, it is an icp warning.

shay-berman: remember to comment imagePullSecrets it out and don't put "secret1" just secretkey.

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
shay-berman: agree with Matt
please add some comment to givre context

shay-berman:
LGTM
make sure it pass CI staging and merge to develop

27149chen:
> > For livenessProbe I think the same version should be used by both for v1. We don't want different versions to accidentally be used at this time.
> > Global imagePullSecret and pullPolicy would be a good features for v1.
> > nodeSelector option should be exposed for node:. There may be cases were not all worker nodes are storage attached.
> 
> Agree with Matt
> 
>     1. imagePullSecret and pullPolicy - i think its mandatory. (does it for openshift?) - can you add it tomorrow?
> 
>     2. nodeSelector - I think more tolaration then nodeselector. - @27149chen  how much time take to add it? can we do both tomorrow morning and pass staging?

1. done.
2. added, will try to pass staging tomorrow

shay-berman:
@27149chen  do you wait for some feedback before merging?

matt-levan:
I tested this PR with OLM in OCP 4.2 and it worked for me.

PR #29, opened 2019-11-07
remove liveness probe from sidecars

27149chen:
Signed-off-by: 27149chen <alex1988@outlook.com>


oriyarde:

PR #35, opened 2019-11-14
Feature/csi 706 csi driver in 1.13

27149chen:
1. add csi driver support for 1.13
2. add related roles to support non kube-system ns
3. kubernetes.io/arch => beta.kubernetes.io/arch

### DIFF
--- a/deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
+++ b/deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
@@ -19,7 +19,7 @@ spec:
         requiredDuringSchedulingIgnoredDuringExecution:
           nodeSelectorTerms:
             - matchExpressions:
-                - key: kubernetes.io/arch
+                - key: beta.kubernetes.io/arch
                   operator: In
                   values:
                     - amd64
@@ -35,7 +35,7 @@ spec:
         requiredDuringSchedulingIgnoredDuringExecution:
           nodeSelectorTerms:
             - matchExpressions:
-                - key: kubernetes.io/arch
+                - key: beta.kubernetes.io/arch
                   operator: In
                   values:
                     - amd64
--- a/pkg/config/resources.go
+++ b/pkg/config/resources.go
@@ -37,6 +37,10 @@ const (
 	ExternalAttacherClusterRoleBinding    ResourceName = "external-attacher-clusterrolebinding"
 	ExternalSnapshotterClusterRole        ResourceName = "external-snapshotter-clusterrole"
 	ExternalSnapshotterClusterRoleBinding ResourceName = "external-snapshotter-clusterrolebinding"
+	SCCForControllerClusterRole           ResourceName = "scc-controller"
+	SCCForControllerClusterRoleBinding    ResourceName = "scc-controller"
+	SCCForNodeClusterRole                 ResourceName = "scc-node"
+	SCCForNodeClusterRoleBinding          ResourceName = "scc-node"
 )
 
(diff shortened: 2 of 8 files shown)


deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
```diff
@@ -19,7 +19,7 @@ spec:
         requiredDuringSchedulingIgnoredDuringExecution:
           nodeSelectorTerms:
             - matchExpressions:
-                - key: kubernetes.io/arch
+                - key: beta.kubernetes.io/arch
```
shay-berman: r u sure we will not encounter issue with cert here?
i remember that one of the cert (maybe the ibm one) failed onit.

anyway if redhat cert pass i am ok with it.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -283,15 +302,41 @@ func (r *ReconcileIBMBlockCSI) reconcileCSIDriver(instance *ibmblockcsi.IBMBlock
 			return err
 		}
 	} else if err != nil {
-		recLogger.Error(err, "Failed to get CSIDriver", "Name", cd.GetName())
-		return err
+		if r.serverVersion == "1.13" {
+			return r.reconcileCSIDriver113(instance)
+		} else {
+			recLogger.Error(err, "Failed to get CSIDriver", "Name", cd.GetName())
```
shay-berman: r u sure here?
why to fail of no 1.13?
iguess if ci pass it means that its ok
but just double check

shay-berman:
suggestion

since its only for openshift side of thing
then we can publish operator in redhat registry with 1.0.0 tag.    so its ok u didnt change to 1.0.1.

so from the operatorhub it still tagged 1.0.0  which is ok.   so we keep the same version all the way

thoughts?

deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr.yaml
matt-levan: IBM cert checks for beta.kubernetes.io/arch.  So no issue there.  Will it cause issues with newer k8s version though (even though not supported)?

pkg/config/resources.go
```diff
@@ -37,6 +37,10 @@ const (
 	ExternalAttacherClusterRoleBinding    ResourceName = "external-attacher-clusterrolebinding"
 	ExternalSnapshotterClusterRole        ResourceName = "external-snapshotter-clusterrole"
 	ExternalSnapshotterClusterRoleBinding ResourceName = "external-snapshotter-clusterrolebinding"
+	SCCForControllerClusterRole           ResourceName = "scc-controller"
+	SCCForControllerClusterRoleBinding    ResourceName = "scc-controller"
+	SCCForNodeClusterRole                 ResourceName = "scc-node"
+	SCCForNodeClusterRoleBinding          ResourceName = "scc-node"
```
matt-levan: You are not creating an SCC here, but assign clusterRole and ClusterRoleBidnings, correct?  The Resource names should be:
<serviceAccount>-<node/controller>-<clusterrole/clusterrolebinding>

Maybe the funtion name should be AddSCCTo<role> as that would map more closely to the oc adm command.
oc adm policy add-scc-to-user

matt-levan:
Since we GA today with version 1.0.0.  Should we update Operator to 1.0.1 for both docker hub and RH registry in case anyone has already pulled the version from Docker?

shay-berman:
hi

so great pr merged

i see that u keep v1.0.0

oriyarde:

PR #38, opened 2019-12-10
Minor formatting proposal on paragraph 64

kant:


27149chen:
@kant , thank you for your improvement, but according to company policy, we are not ready to accept outside contribution yet, I'm very sorry for that.

kant:
@27149chen Do not worry, I do for for the sake of proper understanding of the documentation (the less cumbersome, the better)

PR #39, opened 2019-12-23
update crds in make update

27149chen:
update code generation for crd changes in make update


hack/update-crds.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+
+# Run operator-sdk generate k8s and operator-sdk generate openapi to update code after crd changes.
+
+if ! [ -x "$(command -v operator-sdk)" ]; then
+  echo 'Error: operator-sdk is not installed.' >&2
+  exit 1
+fi
+
+echo "run operator-sdk generate k8s"
```
oriyarde: consider using "set -x" instead

oriyarde: by "instead" i meant that there is no need for "echo"s

27149chen: can you explain why?

oriyarde: since "-x" already echoes the commands

27149chen: ok, got you.

PR #46, opened 2020-02-13
Story/csi 1169 ocp4 3 operator deployment

igorgonibm:
add  patch permission to csi-attacher


oriyarde:
@igorgonibm could we delete this branch?

PR #54, opened 2020-03-13
generate release and version in dockerfile

27149chen:


oriyarde:
@27149chen could you do the same in the driver?

27149chen:
@oriyarde could you do it, if not, I'll do it next week, I am stuck in a ds8k problem.

oriyarde:
@27149chen since both tickets are currently assigned to you, please consult with @erantzabari regarding their priorities and assignments

PR #55, opened 2020-03-15
update readme

27149chen:
update readme and csv


oriyarde:
I think we should wait for the 1.1.0 User Guide to be finished before updating its duplications (e.g. README files)

README.md
```diff
@@ -228,6 +321,9 @@ storageclass.storage.k8s.io/gold created
 <br/>
 <br/>
 
+## Upgrading
+
+In order to upgrade the CSI operator and driver from a previous version, uninstall the existing driver and then install the newer version.
```
27149chen: We do not support upgrading, we should tell the user about it, please update the words here and change it in UG accordingly

oriyarde: I think it's technically upgrading, since the result is the same. it's just not seamless...

rivkap: I agree with @oriyarde  - it is still the way to upgrade the version. to say that we don't support upgrade is not a good way to word it to customers who might already be using 1.0.0.
I explain that it is to uninstall and then reinstall. that should be more than sufficient. Note that in the UG it explains how to manually upgrade (as well as how to make it automatic) via OCP.

oriyarde: automatic? this specifically might be something to remove from the UG

rivkap: @oriyarde  @27149chen  please see the UG and provide any comments there, if needed.

oriyarde: done

README.md
```diff
@@ -5,18 +5,22 @@ This is the official operator to deploy and manage IBM block storage CSI driver.
...
   - Kubernetes v1.14
+  - Kubernetes v1.16
 
 Supported IBM storage systems:
   - IBM FlashSystem 9100
-  - IBM Spectrum Virtualize
-  - IBM Storwize
+  - IBM Spectrum Virtualize Family (including IBM Flash family members built with IBM Spectrum Virtualize (FlashSystem 5010, 5030, 5100, 7200, 9100, 9200, 9200R) and IBM SAN Volume Controller (SVC) models SV2, SA2)
   - IBM FlashSystem A9000/R
+  - IBM DS8880
```
rivkap: Instead of these two lines, according to the latest RN - please have this as one line stating "IBM DS8000 Family"

README.md
```diff
@@ -5,18 +5,22 @@ This is the official operator to deploy and manage IBM block storage CSI driver.
...
+  - IBM DS8900
```
rivkap: remove per comment above

README.md
```diff
@@ -52,36 +131,24 @@ systemctl status multipathd
...
-3.3.3. Verify that the login was successful and display all targets that you logged into. The portal value must be the iSCSI target IP address.
-
-```bash
-$ iscsiadm -m session --rescan
-Rescanning session [sid: 1, target: {storage system IQN},
-portal: {STORAGE-SYSTEM-iSCSI-PORT-IP1},{port number}
-portal: {STORAGE-SYSTEM-iSCSI-PORT-IP2},{port number}
-```
+#### 3. Configure storage system connectivity
+3.1. Define the hostname of each Kubernetes node on the relevant storage systems with the valid WWPN(for Fibre Channel) or IQN(for iSCSI) of the node.
```
rivkap: 1. delete "Kubernetes"
Define the hostname of each node on the ...
2. valid WWPN (for ...)
<needs a space between "WWPN" and "(for">
3. or IQN(for ...)
<needs a space between "IQN" and "(for">

README.md
```diff
@@ -5,18 +5,22 @@ This is the official operator to deploy and manage IBM block storage CSI driver.
 
 Supported container platforms:
```
igorgonibm: All the changes should be consistent with driver readme

igorgonibm: All the changes should be consistent with driver readme.

oriyarde: Since the operator readme was reviewed more (and first) and is more important for users, I think the driver readme should be aligned with the operator, and not vice versa.

Regardless, they are not completely identical in 1.0.0, so if something is not aligned, you need to comment on the relevant line (this line looks valid to me)

oriyarde:

PR #58, opened 2020-03-23
upgrade driver during operator upgrade

27149chen:
Per the suggestion from RedHat, we should update the cr during upgrading operator so that the driver will be upgraded together.
This pr
1. Update the controller, node and sidecars images if the defualt values were used in previous version
2. Skip the update if they were modified, for example, a private registry is used.

Jira ticket: CSI-1284

### DIFF
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -42,8 +42,8 @@ const (
 	ENVKubeVersion    = "KUBE_VERSION"
 	ENVPlatform       = "PLATFORM"
 
-	Controller             = "csi-controller"
-	Node                   = "csi-node"
+	ControllerImage        = "csi-controller"
+	NodeImage              = "csi-node"
 	CSINodeDriverRegistrar = "csi-node-driver-registrar"
 	CSIProvisioner         = "csi-provisioner"
 	CSIAttacher            = "csi-attacher"
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -32,8 +32,8 @@ const (
 	NodeRepository       = "ibmcom/ibm-block-csi-driver-node"
 	NodeAgentRepository  = "ibmcom/ibm-node-agent"
 
-	OpenShiftControllerRepository = "ibmcom/ibm-block-csi-controller-driver"
-	OpenShiftNodeRepository       = "ibmcom/ibm-block-csi-node-driver"
+	OpenShiftControllerRepository = "ibmcom/ibm-block-csi-driver-controller"
+	OpenShiftNodeRepository       = "ibmcom/ibm-block-csi-driver-node"
 	OpenShiftNodeAgentRepository  = "ibmcom/ibm-node-agent"
 
 	ENVIscsiAgentPort = "ISCSI_AGENT_PORT"
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -18,6 +18,8 @@ package config
 
 // Add a field here if it never changes, if it changes over time, put it to settings.go
 const (
+	OpenShift   = "OpenShift"
+	Kubernetes  = "Kubernetes"
 	APIGroup    = "csi.ibm.com"
 	APIVersion  = "v1"
 	Name        = "ibm-block-csi-operator"
@@ -30,11 +32,18 @@ const (
 	NodeRepository       = "ibmcom/ibm-block-csi-node-driver"
 	NodeAgentRepository  = "ibmcom/ibm-node-agent"
 
+	OpenShiftControllerRepository = "ibmcom/ibm-block-csi-controller-driver"
+	OpenShiftNodeRepository       = "ibmcom/ibm-block-csi-node-driver"
+	OpenShiftNodeAgentRepository  = "ibmcom/ibm-node-agent"
+
 	ENVIscsiAgentPort = "ISCSI_AGENT_PORT"
 	ENVEndpoint       = "ENDPOINT"
 	ENVNodeName       = "NODE_NAME"
 	ENVKubeVersion    = "KUBE_VERSION"
+	ENVPlatform       = "PLATFORM"
 
+	Controller             = "csi-controller"
+	Node                   = "csi-node"
 	CSINodeDriverRegistrar = "csi-node-driver-registrar"
 	CSIProvisioner         = "csi-provisioner"
 	CSIAttacher            = "csi-attacher"
--- a/pkg/config/constants.go
+++ b/pkg/config/constants.go
@@ -28,8 +28,8 @@ const (
 	DeployPath  = "/deploy"
 	Masterlabel = "node-role.kubernetes.io/master"
 
-	ControllerRepository = "ibmcom/ibm-block-csi-controller-driver"
-	NodeRepository       = "ibmcom/ibm-block-csi-node-driver"
+	ControllerRepository = "ibmcom/ibm-block-csi-driver-controller"
+	NodeRepository       = "ibmcom/ibm-block-csi-driver-node"
 	NodeAgentRepository  = "ibmcom/ibm-node-agent"
 
 	OpenShiftControllerRepository = "ibmcom/ibm-block-csi-controller-driver"
--- a/pkg/config/settings.go
+++ b/pkg/config/settings.go
@@ -16,11 +16,18 @@
 
 package config
 
+import "k8s.io/apimachinery/pkg/util/sets"
+
 const (
-	NodeDriverRegistrarImage    = "quay.io/k8scsi/csi-node-driver-registrar:v1.2.0"
-	CSIProvisionerImage         = "quay.io/k8scsi/csi-provisioner:v1.3.0"
-	CSIAttacherImage            = "quay.io/k8scsi/csi-attacher:v1.2.1"
-	CSILivenessProbeImage       = "quay.io/k8scsi/livenessprobe:v1.1.0"
+	NodeDriverRegistrarImage = "quay.io/k8scsi/csi-node-driver-registrar:v1.2.0"
+	CSIProvisionerImage      = "quay.io/k8scsi/csi-provisioner:v1.4.0"
+	CSIAttacherImage         = "quay.io/k8scsi/csi-attacher:v1.2.1"
+	CSILivenessProbeImage    = "quay.io/k8scsi/livenessprobe:v1.1.0"
+
+	OpenShiftNodeDriverRegistrarImage = "registry.redhat.io/openshift4/ose-csi-driver-registrar:v4.3"
+	OpenShiftCSIProvisionerImage      = "registry.redhat.io/openshift4/ose-csi-external-provisioner-rhel7:v4.3"
+	OpenShiftCSIAttacherImage         = "registry.redhat.io/openshift4/ose-csi-external-attacher:v4.3"
+	OpenShiftCSILivenessProbeImage    = "registry.redhat.io/openshift4/ose-csi-livenessprobe:v4.3"
 
 	ControllerTag = "1.1.0"
 	NodeTag       = "1.1.0"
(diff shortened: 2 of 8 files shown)


oriyarde:
@27149chen what if we want the operator to be supported on different platform versions, but with different sidecar versions (e.g. [provisioner 1.4.0 on k8s 1.1.6 but provisioner 1.5.0 on k8s 1.17](https://kubernetes-csi.github.io/docs/external-provisioner.html))?

27149chen:
we didit before for 1.13 and 1.14. what is the difference between these two versions? are they incompatible？

pkg/config/constants.go
```diff
@@ -30,11 +32,18 @@ const (
 	NodeRepository       = "ibmcom/ibm-block-csi-node-driver"
 	NodeAgentRepository  = "ibmcom/ibm-node-agent"
 
+	OpenShiftControllerRepository = "ibmcom/ibm-block-csi-controller-driver"
```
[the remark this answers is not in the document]
27149chen: csv examples varies with different versions, I don't think it is possible. Even it is possible, fields in the csv examples should come from the code. Maybe next time, operator-sdk will give us a tool to geretate csv, we can use that to generate the examples.

pkg/config/constants.go
```diff
@@ -30,11 +32,18 @@ const (
...
+	OpenShiftNodeRepository       = "ibmcom/ibm-block-csi-node-driver"
+	OpenShiftNodeAgentRepository  = "ibmcom/ibm-node-agent"
```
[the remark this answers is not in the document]
27149chen: yes, but it is disabled by default.

pkg/config/constants.go
```diff
@@ -30,11 +32,18 @@ const (
...
+
 	ENVIscsiAgentPort = "ISCSI_AGENT_PORT"
 	ENVEndpoint       = "ENDPOINT"
 	ENVNodeName       = "NODE_NAME"
 	ENVKubeVersion    = "KUBE_VERSION"
+	ENVPlatform       = "PLATFORM"
 
+	Controller             = "csi-controller"
```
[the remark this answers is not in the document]
27149chen: Maybe it is a bit confusing. But one stands for the image and the other stands for a resource, they are different.

pkg/internal/ibmblockcsi/default_setter.go
```diff
@@ -17,32 +17,109 @@
...
-		c.Spec.Node.Repository = config.NodeRepository
-		c.Spec.Node.Tag = config.NodeTag
+		regAndTag := strings.Split(c.GetDefaultImageByName(platform, config.Node), ":")
+		c.Spec.Node.Repository = regAndTag[0]
+		c.Spec.Node.Tag = regAndTag[1]
 
 		changed = true
 	}
 
+	if ch := c.replaceSidecars(platform); ch {
```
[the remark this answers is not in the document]
27149chen: I think it is clear enough.

27149chen:
> > we didit before for 1.13 and 1.14. what is the difference between these two versions? are they incompatible？
> 
> I don't know if my specific example has incompatibility (though it's certainly recommended), but my question is also relevant for incompatibility scenario.

Let's haddle it like the way we did for 1.13 and 1.14 when we need it.

27149chen:
> > Per the suggestion from RedHat, we should update the cr during upgrading operator so that the driver will be upgraded together.
> > This pr
> > 
> > 1. Update the controller, node and sidecars images if the defualt values were used in previous version
> > 2. Skip the update if they were modified, for example, a private registry is used.
> 
> @27149chen I couldn't access the ticket - could you add us to the ticket/share the RedHat suggestion info here/wiki?

Added in the ticket.

pkg/internal/ibmblockcsi/ibmblockcsi.go
```diff
@@ -119,3 +122,86 @@ func (c *IBMBlockCSI) GetCSINodeImage() string {
...
+
+func (c *IBMBlockCSI) GetSidecarImageByName(name string) string {
+	sidecar := c.GetSidecarByName(name)
+	if sidecar != nil {
+		return fmt.Sprintf("%s:%s", sidecar.Repository, sidecar.Tag)
+	}
+	return ""
+}
+
+func (c *IBMBlockCSI) GetDefaultImageByName(platform, name string) string {
```
igorgonibm: GetDefaultContainerImageByPlatformAndName

pkg/internal/ibmblockcsi/ibmblockcsi.go
```diff
@@ -119,3 +122,86 @@ func (c *IBMBlockCSI) GetCSINodeImage() string {
...
+	case config.OpenShift:
+		return c.GetOpenShiftDefaultImageByName(name)
+	case config.Kubernetes:
+		return c.GetKubernetesDefaultImageByName(name)
+	default:
+		return ""
+	}
+}
+
+func (c *IBMBlockCSI) GetKubernetesDefaultImageByName(name string) string {
```
igorgonibm: GetKubernetesDefaultContainerImageByName

pkg/internal/ibmblockcsi/ibmblockcsi.go
```diff
@@ -119,3 +122,86 @@ func (c *IBMBlockCSI) GetCSINodeImage() string {
...
+	case config.CSINodeDriverRegistrar:
+		return config.NodeDriverRegistrarImage
+	case config.LivenessProbe:
+		return config.CSILivenessProbeImage
+	default:
+		return ""
+	}
+}
+
+func (c *IBMBlockCSI) GetOpenShiftDefaultImageByName(name string) string {
```
igorgonibm: GetOpenShiftDefaultContainerImageByName

pkg/internal/ibmblockcsi/default_setter.go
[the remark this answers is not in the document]
oriyarde: `ch` is definitely not clear enough

pkg/internal/ibmblockcsi/ibmblockcsi.go
27149chen: added platform, but I don't want to add container

pkg/internal/ibmblockcsi/ibmblockcsi.go
27149chen: I don't want to add "container"

pkg/internal/ibmblockcsi/ibmblockcsi.go
27149chen: I don't want to add "container"

pkg/config/constants.go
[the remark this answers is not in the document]
oriyarde: so maybe rename to ControllerContainerName

pkg/config/settings.go
```diff
@@ -32,3 +39,86 @@ const (
...
+var ReplaceOpenShiftControllerVersions = sets.String{
+	OpenShiftControllerRepository + ":" + "1.0.0": sets.Empty{},
+}
+
+var ReplaceOpenShiftNodeVersions = sets.String{
+	OpenShiftNodeRepository + ":" + "1.0.0": sets.Empty{},
+}
+
+var ReplaceOpenShiftCSIProvisionerVersions = sets.String{
+	"quay.io/k8scsi/csi-provisioner:v1.3.0": sets.Empty{},
```
oriyarde: these should come from consts

pkg/internal/ibmblockcsi/default_setter.go
```diff
@@ -17,32 +17,109 @@
...
+	}
+	if ch := c.setDefualtSidecars(platform); ch {
+		changed = ch
+	}
+
+	return changed
+}
+
+// replacePreviousVersion replaces the previous version of controller and node
+// images to new version during upgrade.
```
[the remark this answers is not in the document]
oriyarde: could respond here before resolve?

pkg/internal/ibmblockcsi/ibmblockcsi.go
igorgonibm: I don't see the new name. Was it pushed?

pkg/internal/ibmblockcsi/default_setter.go
```diff
@@ -17,32 +17,109 @@
...
+	if config.GetReplaceVersions(platform, config.Node).Has(c.GetCSINodeImage()) {
+		c.Spec.Node.Repository = ""
+		c.Spec.Node.Tag = ""
 		changed = true
 	}
 
 	return changed
 }
+
+func (c *IBMBlockCSI) clearSidecars(platform string) bool {
```
oriyarde: `clearSidecarsPreviousVersions`?

pkg/config/constants.go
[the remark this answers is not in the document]
27149chen: ok, rename it as ControllerImage

pkg/config/settings.go
27149chen: ReplaceOpenShiftCSIProvisionerVersions is also a const

pkg/internal/ibmblockcsi/default_setter.go
27149chen: ok, rename to clearSidecarPreviousVersion

pkg/config/settings.go
oriyarde: then I'm not sure I follow - what happens if in CSI 1.2.0 we upgrade our provisioner to 1.5.0? do we need both 1.3.0 and 1.4.0 here? where is the version comparison?

27149chen: yes, both 1.3 and 1.4.
What do you mean by version comparison?

oriyarde: I thought there might be missing some version comparison logic between current and latest version, but I guess new CSI versions want the option to also downgrade sidecars in the future.

I still don't like the idea of having to add each used version to a set (how will the maintain even know to do it?).
I still think it should come from the OLM directory, all the info is already there, and if the `alm-examples` format will change, we will add parsing support for it here.

oriyarde: This is exactly the opposite of CSI-1055 purpose

27149chen: 1. we do not support downgrade, if the user use an old image in new version, wew can not guarantee the availability.
2. I think it is part of the upgrade procedure (replacing old image with new one), and it is easy to understand. Don't worry you will forget it, we have test cases to cover it.
3. I still think the code is the source, the examples should come from the code in some way.
4. Sorry, I don't understand CSI-1055, there is no context.

27149chen:
> @27149chen I think we need to reconsider if this is for 1.1.0.
> We don't even have test automation for upgrade.

We tested the upgrade in ocp 4.2 several times, both auto and manual. Please ask first if you don't know.

oriyarde:
> > @27149chen I think we need to reconsider if this is for 1.1.0.
> > We don't even have test automation for upgrade.
> 
> We tested the upgrade in ocp 4.2 several times, both auto and manual.

I did not review any test automation code for this. The test cases are complex, please refer us to the relevant automation.

pkg/internal/ibmblockcsi/default_setter.go
oriyarde: I meant sidecars (plural), unless I misunderstood the function

pkg/config/constants.go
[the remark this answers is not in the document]
oriyarde: By "Image" do you mean a docker image? because AFAIK this is the name of the container inside the pod

pkg/config/constants.go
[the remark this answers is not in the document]
oriyarde: I don't think it should come from the code. the info is already in `alm-examples`, saved forever by version, which is exactly what we need here.

pkg/config/settings.go
oriyarde: 1. I didn't mean the user choose to use an old image, I meant that we want the option to officially downgrade a sidecar in a new CSI release.
2. I don't think it's easy for the user to know exactly which old images we decide to replace or not, since they shouldn't be aware of our entire versions history.
3. If you know of a way to generate the examples from the code, that will also be fine.
4. The purpose of CSI-1055 is to reduce the amount of places where we maintain the images and versions, ideally to one location (this PR currently does the opposite).

oriyarde:
> > > @27149chen I think we need to reconsider if this is for 1.1.0.
> > > We don't even have test automation for upgrade.
> > 
> > 
> > We tested the upgrade in ocp 4.2 several times, both auto and manual.
> 
> I did not review any test automation code for this. The test cases are complex, please refer us to the relevant automation.

@27149chen I have confirmed with Ke Xu, as I wrote, we don't have any automation tests for it.

PR #61, opened 2020-03-26
change from "Seamless Upgrades" to "Basic Install"

27149chen:
change capability from "Seamless Upgrades" to "Basic Install"


oriyarde:

PR #62, opened 2020-03-29
increase sidecar timeout interval to 30s

27149chen:
increase provisioner and attacher sidecar timeout interval to 30s


oriyarde:

PR #68, opened 2020-04-22
upgrade/CSI-1373 operator ubi tag

oriyarde:
https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi7/ubi-minimal
should improve our health index from D to A

### DIFF
--- a/build/Dockerfile.operator
+++ b/build/Dockerfile.operator
@@ -24,7 +24,7 @@ RUN CGO_ENABLED=1 GOOS=linux go build \
 	github.com/IBM/ibm-block-csi-operator/cmd/manager
 
 
-FROM registry.access.redhat.com/ubi7/ubi-minimal:7.7-138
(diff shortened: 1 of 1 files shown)


build/Dockerfile.operator
```diff
@@ -24,7 +24,7 @@ RUN CGO_ENABLED=1 GOOS=linux go build \
 	github.com/IBM/ibm-block-csi-operator/cmd/manager
 
 
-FROM registry.access.redhat.com/ubi7/ubi-minimal:7.7-138
+FROM registry.access.redhat.com/ubi7/ubi-minimal:7.8-237
```
igorgonibm, line 27: why the change?

oriyarde: our Red Hat image health index is down to D:
https://access.redhat.com/containers/?tab=tags#/registry.connect.redhat.com/ibm/ibm-block-csi-operator
probably for the same reasons that our base image has it.
so this should fix it

PR #75, opened 2020-06-09
Story/csi 1263 connection pool

igorgonibm:
Increase liveness probe timeout (how much time till it thinks controller is down)


oriyarde:
@igorgonibm could you add a description for this PR?
it's not clear to me what `3` (/`20`) means here

PR #88, opened 2020-07-11
add support for Z and Power installation

oriyarde:


deploy/crds/csi.ibm.com_v1_ibmblockcsi_cr_ocp.yaml
```diff
@@ -0,0 +1,67 @@
...
+                    - s390x
+                    - ppc64le
+
+  sidecars:
+  - name: csi-node-driver-registrar
+    repository: registry.redhat.io/openshift4/ose-csi-driver-registrar
+    tag: v4.3
+    imagePullPolicy: IfNotPresent
+  - name: csi-provisioner
+    repository: registry.redhat.io/openshift4/ose-csi-external-provisioner-rhel7
```
ArbelNathan, line 53: as I understand from the file name the yaml is for installation on ocp, its not a problem that the repository used here is named with rhel7?

oriyarde: It's not that I agree with their naming, but I couldn't find a version without it
https://github.com/IBM/ibm-block-csi-operator/blob/7b095f15c347c4dad2c8eaf7b0b2f62ee042ce28/deploy/olm-catalog/ibm-block-csi-operator/ibm-block-csi-operator.v1.2.0.clusterserviceversion.yaml#L83

ArbelNathan: ok

PR #94, opened 2020-09-03
CSI-1737 advance version to 1.3.0

oriyarde:
similar PR in driver: https://github.com/IBM/ibm-block-csi-driver/pull/222


version/version.go
```diff
@@ -17,6 +17,6 @@
 package version
 
 var (
-	Version       = "1.2.0"
-	DriverVersion = "1.2.0"
+	Version       = "1.3.0"
+	DriverVersion = "1.3.0"
```
ArbelNathan, line 21: what the diff?

oriyarde: what do you mean?

ArbelNathan: version vs DriverVersion

oriyarde: [your guess is as good as mine](https://github.com/IBM/ibm-block-csi-operator/pull/76/files#r438116211).
`DriverVersion` sounds pretty straightforward, and I see `Version` usage [here](https://github.com/IBM/ibm-block-csi-operator/blob/aa140208f60135ad1e7819bb85ce53bfe63d8f8d/pkg/internal/ibmblockcsi/ibmblockcsi.go#L72)

ArbelNathan: so Driver is controller an Node? like in the github?

ArbelNathan: ok, sounds right, if we will have patches it will be handy to separate it..

ArbelNathan (approved):
LGTM

barakda (approved):

PR #110, opened 2020-11-24
CSI-1860 bump base image version

oriyarde:
similar to https://github.com/IBM/ibm-block-csi-operator/pull/68


build/Dockerfile.operator
```diff
@@ -24,7 +24,7 @@ RUN CGO_ENABLED=1 GOOS=linux go build \
 	github.com/IBM/ibm-block-csi-operator/cmd/manager
 
 
-FROM registry.access.redhat.com/ubi7/ubi-minimal:7.8-237
+FROM registry.access.redhat.com/ubi7/ubi-minimal:7.9-224
```
ArbelNathan, line 27: why not `latest`?

oriyarde: `latest` is not deterministic.
if you find any `latest` in our projects, please let me know

barakda (approved):

build/Dockerfile.operator
ArbelNathan: what do you mean by "not deterministic"?

oriyarde: with `latest`, building the same commit at different times could yield different outputs - which is bad for CI stability

ArbelNathan (approved):

PR #132, opened 2021-02-21
CSI-2777 fix driver stuck on creating after token change

oriyarde:
based on: https://docs.openshift.com/container-platform/4.6/nodes/pods/nodes-pods-secrets.html#nodes-pods-secrets-updating_nodes-pods-secrets
and: https://stackoverflow.com/a/59051313

### DIFF
--- a/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
+++ b/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
@@ -259,16 +259,14 @@ func (r *ReconcileIBMBlockCSI) updateStatus(instance *ibmblockcsi.IBMBlockCSI, o
 	if instance.Status.ControllerReady && instance.Status.NodeReady {
 		phase = csiv1.DriverPhaseRunning
 	} else {
-		if originalStatus.Phase == csiv1.DriverPhaseRunning {
-			key, value := r.getRolloutRestartAnnotation()
-			if !instance.Status.ControllerReady {
-				controller.Spec.Template.ObjectMeta.Annotations[key] = value
-				controllerRolloutRestart = true
-			}
-			if !instance.Status.NodeReady {
-				node.Spec.Template.ObjectMeta.Annotations[key] = value
-				nodeRolloutRestart = true
-			}
+		key, value := r.getRolloutRestartAnnotation()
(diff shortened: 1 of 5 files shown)


pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -263,15 +259,53 @@ func (r *ReconcileIBMBlockCSI) updateStatus(instance *ibmblockcsi.IBMBlockCSI) e
 	if instance.Status.ControllerReady && instance.Status.NodeReady {
 		phase = csiv1.DriverPhaseRunning
 	} else {
+		key, value := r.getRolloutRestartAnnotation()
```
zingero, line 262: I suggest renaming the key/value to meaningful names such as:
restartedAt, timestamp

oriyarde: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -263,15 +259,69 @@ func (r *ReconcileIBMBlockCSI) updateStatus(instance *ibmblockcsi.IBMBlockCSI) e
...
+	if !reflect.DeepEqual(originalStatus, instance.Status) {
+		reqLogger.Info("updating IBMBlockCSI status", "name", instance.Name, "from", originalStatus, "to", instance.Status)
+		sErr := r.client.Status().Update(context.TODO(), instance.Unwrap())
+		if sErr != nil {
+			return sErr
+		}
+	}
+
+	if controllerRestart {
+		reqLogger.Info("restarting controller")
```
ArbelNathan, line 282: "due to status change"? I think more info is needed.. if all we see in logger is restarting it can be confusing (especially for users)

ArbelNathan: in node too

oriyarde: done

ArbelNathan (approved):
LGTM

zingero (approved):

PR #133, opened 2021-02-24
CSI-2716 add cluster roles + bindings finalizer

oriyarde:
k8s 1.20 removed garbage collection of our cluster roles + bindings, as can be seen by the warnings shown in the following command:
`kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace`

as explained [here](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/):

> In v1.20+, if the garbage collector detects ... a cluster-scoped dependent with an ownerReference referencing a namespaced kind, a warning Event ... is reported.

and [here](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#urgent-upgrade-notes):

> A cluster-scoped object with an ownerReference referencing a uid of a namespaced kind is now consistently treated as though that owner is not resolvable, and the child object is ignored by the garbage collector.

in our case the cluster roles/rolebindings each have an `ownerReference` to our namespaced CR (set by the operator [here](https://github.com/IBM/ibm-block-csi-operator/blob/dff4aadfd2d2fefa466e4d0573a8cd6c75aadd62/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go#L357) and [here](https://github.com/IBM/ibm-block-csi-operator/blob/dff4aadfd2d2fefa466e4d0573a8cd6c75aadd62/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go#L401)), which used to guarantee their cleanup, but not anymore.

we use cluster-level roles+rolebidings to allow the csi-controller service-account access to cluster-level objects such as storageclass, which the sidecars require. 

since the operator and its CR should remain namespaced, for all the benefits of using a namespace, the solution here is to add a finalizer in the operator that deletes the cluster roles+rolebindings when the CR is deleted.

### DIFF
--- a/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
+++ b/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
@@ -207,7 +207,7 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
 		if err != nil {
 			return reconcile.Result{}, err
 		}
-		
+
 		if !isFinalizerExists {
 			return reconcile.Result{}, nil
 		}
--- a/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
+++ b/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
@@ -281,10 +281,7 @@ func (r *ReconcileIBMBlockCSI) hasFinalizer(instance *ibmblockcsi.IBMBlockCSI) (
 		return false, err
 	}
 
-	if util.Contains(accessor.GetFinalizers(), finalizerName) {
-		return true, nil
-	}
-	return false, nil
+	return util.Contains(accessor.GetFinalizers(), finalizerName), nil
 }
 
 func (r *ReconcileIBMBlockCSI) removeFinalizer(instance *ibmblockcsi.IBMBlockCSI) error {
--- a/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
+++ b/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
@@ -24,12 +24,13 @@ import (
 	"strings"
 	"time"
 
-	csiv1 "github.com/IBM/ibm-block-csi-operator/pkg/apis/csi/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1beta1 "k8s.io/api/storage/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -44,8 +45,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	csiv1 "github.com/IBM/ibm-block-csi-operator/pkg/apis/csi/v1"
 	oconfig "github.com/IBM/ibm-block-csi-operator/pkg/config"
 	clustersyncer "github.com/IBM/ibm-block-csi-operator/pkg/controller/ibmblockcsi/syncer"
+	"github.com/IBM/ibm-block-csi-operator/pkg/controller/util"
 	"github.com/IBM/ibm-block-csi-operator/pkg/internal/ibmblockcsi"
 	kubeutil "github.com/IBM/ibm-block-csi-operator/pkg/util/kubernetes"
 	oversion "github.com/IBM/ibm-block-csi-operator/version"
@@ -192,8 +195,32 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
 			err = fmt.Errorf("failed to update IBMBlockCSI CR: %v", err)
 			return reconcile.Result{}, err
 		}
+		return reconcile.Result{}, nil
 	}
 
+	if err := r.addFinalizerIfNotPresent(instance); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if !instance.GetDeletionTimestamp().IsZero() {
+		isFinalizerExists, err := r.hasFinalizer(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		
+		if !isFinalizerExists {
+			return reconcile.Result{}, nil
+		}
+
+		if err := r.deleteClusterRolesAndBindings(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		if err := r.removeFinalizer(instance); err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
 
 	originalStatus := *instance.Status.DeepCopy()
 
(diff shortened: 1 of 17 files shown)


pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -228,8 +255,69 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
...
+	logger := log.WithName("getAccessorAndFinalizerName")
+	lowercaseKind := strings.ToLower(instance.GetObjectKind().GroupVersionKind().Kind)
+	finalizerName := fmt.Sprintf("%s.%s", lowercaseKind, oconfig.APIGroup)
+
+	accessor, err := meta.Accessor(instance)
+	if err != nil {
+		logger.Error(err, "failed to get meta information of instance")
+		return nil, "", err
+	}
+	return accessor, finalizerName, nil
```
zingero, line 316: Is it considered good practice to return a tuple of three elements in Go? IMHO, when we want to return more than two elements, we usually wrap them with a dictionary/object since we don't want to "remember" the order of the elements whenever we call this function.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -228,8 +255,69 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
...
+func (r *ReconcileIBMBlockCSI) removeFinalizer(instance *ibmblockcsi.IBMBlockCSI) error {
+	logger := log.WithName("removeFinalizer")
+
+	accessor, finalizerName, err := r.getAccessorAndFinalizerName(instance)
+	if err != nil {
+		return err
+	}
+
+	accessor.SetFinalizers(util.Remove(accessor.GetFinalizers(), finalizerName))
+	if err := r.client.Update(context.TODO(), instance.Unwrap()); err != nil {
```
zingero, line 299: I suggest breaking this line to two lines:
```suggestion
	err := r.client.Update(context.TODO(), instance.Unwrap())
	if err != nil {
```
since this line does too many things:
1. `if` statement
2. local variable declaration
3. client updating

WDYT?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -228,8 +255,69 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
...
+func (r *ReconcileIBMBlockCSI) hasFinalizer(instance *ibmblockcsi.IBMBlockCSI) (bool, error) {
+	accessor, finalizerName, err := r.getAccessorAndFinalizerName(instance)
+	if err != nil {
+		return false, err
+	}
+
+	if util.Contains(accessor.GetFinalizers(), finalizerName) {
+		return true, nil
+	}
+	return false, nil
```
zingero, line 287: Can we merge these lines together to:
```suggestion
	return util.Contains(accessor.GetFinalizers(), finalizerName), nil
```
?
The function name and this code are pretty clear even without this `if` statement.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -228,8 +255,69 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileIBMBlockCSI) addFinalizerIfNotPresent(instance *ibmblockcsi.IBMBlockCSI) error {
+	logger := log.WithName("addFinalizer")
```
zingero, line 259: Is there any convention for the logger text?
In this case, it's only the function's prefix.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -228,8 +255,69 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
...
+	accessor, finalizerName, err := r.getAccessorAndFinalizerName(instance)
+	if err != nil {
+		return err
+	}
+
+	if !util.Contains(accessor.GetFinalizers(), finalizerName) {
+		logger.Info("adding", "finalizer", finalizerName, "on", accessor.GetName())
+		accessor.SetFinalizers(append(accessor.GetFinalizers(), finalizerName))
+
+		if err := r.client.Update(context.TODO(), instance.Unwrap()); err != nil {
```
zingero, line 270: Same as in https://github.com/IBM/ibm-block-csi-operator/pull/133/files#r584316456

pkg/controller/util/utils.go
```diff
@@ -52,3 +52,21 @@ func IsNodeAgentReady(c client.Client) bool {
...
+			return true
+		}
+	}
+	return false
+}
+
+func Remove(list []string, s string) []string {
+	for i, v := range list {
+		if v == s {
+			list = append(list[:i], list[i+1:]...)
```
zingero, line 68: Does this function removes the first `s` or all of the occurrences of `s`?
If the answer is only the first occurrence, then we can break the loop here.
Otherwise, I'm not sure how Go iterates over a list while changing it.

pkg/controller/util/utils.go
```diff
@@ -52,3 +52,21 @@ func IsNodeAgentReady(c client.Client) bool {
...
+		}
+	}
+	return list
+}
```
zingero, line 72: Doesn't Go offer an implementation for this logic?

Anyway, I'm not sure these functions are relevant to this file.
I suggest extracting them to another file i.e. `pkg/controller/util/datastructures.go`

pkg/controller/util/utils.go
```diff
@@ -52,3 +52,21 @@ func IsNodeAgentReady(c client.Client) bool {
 	}
 	return conf.Status.NodeAgent.Phase == csiv1.NodeAgentPhaseRunning
 }
+
+func Contains(list []string, s string) bool {
```
zingero, line 56: Is it possible to make these functions generic so if we want to remove an integer from an integers list, we don't need to implement a similar function?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -192,8 +195,32 @@ func (r *ReconcileIBMBlockCSI) Reconcile(request reconcile.Request) (reconcile.R
 			err = fmt.Errorf("failed to update IBMBlockCSI CR: %v", err)
 			return reconcile.Result{}, err
 		}
+		return reconcile.Result{}, nil
```
zingero, line 198: Adding this `return` statement means changing the software behavior since it won't execute lines 225-255.
Why do we want to return here?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -369,88 +457,114 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+	clusterRoles := r.getClusterRoles(instance)
+
+	for _, cr := range clusterRoles {
+		found := &rbacv1.ClusterRole{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{
+			Name:      cr.Name,
+			Namespace: cr.Namespace,
+		}, found)
+		if err != nil && errors.IsNotFound(err) {
+			continue
```
zingero, line 529: Why don't we want to log the error in case we're not familiar with the error?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -369,88 +457,114 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
 		nodeSCC,
-	} {
-		if err := controllerutil.SetControllerReference(instance.Unwrap(), crb, r.scheme); err != nil {
-			return err
-		}
+	}
+}
+
+func (r *ReconcileIBMBlockCSI) reconcileClusterRoleBinding(instance *ibmblockcsi.IBMBlockCSI) error {
+	recLogger := log.WithValues("Resource Type", "ClusterRoleBinding")
```
zingero, line 563: What does the "rec" mean in `recLogger`? Short for reconcile? recursive? recording?
I suggest removing it as in other logger variables or writing the full word.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -472,3 +586,48 @@ func (r *ReconcileIBMBlockCSI) reconcileClusterRoleBinding(instance *ibmblockcsi
...
+	clusterRoleBindings := r.getClusterRoleBindings(instance)
+
+	for _, crb := range clusterRoleBindings {
+		found := &rbacv1.ClusterRoleBinding{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{
+			Name:      crb.Name,
+			Namespace: crb.Namespace,
+		}, found)
+		if err != nil && errors.IsNotFound(err) {
+			continue
```
zingero, line 602: Why don't we want to log the error in case we're not familiar with the error?

zingero (changes requested):
I suggest having a meeting before changing the code.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: what is special about the number "two" in this rule? and how would you name the object here?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: it's just a common syntax suger, both look ok to me

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: I don't think we have an official convention, though in my current change it's usually the full name of the function.
do you think it would be better to include its full name here?

pkg/controller/util/utils.go
oriyarde: not that I know of

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: we are familiar with the error, it is a "not found" error here

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: see [previous conversation](https://github.com/IBM/ibm-block-csi-operator/pull/133#discussion_r586423233)

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: it was part of develop branch, not relevant to this PR, fixed

pkg/controller/util/utils.go
oriyarde: it removed only the first occurrence, but I improved it to remove all occurrences

pkg/controller/util/utils.go
oriyarde: not that I know of.
done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: removed

ArbelNathan (approved):
LGTM

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: done

zingero: I think so. Otherwise, it may confuse the log readers.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
zingero: Ohhh. I thought the error wasn't found :)

zingero (approved):

PR #135, opened 2021-03-03
license add (CSI-2484)

barakda:


ArbelNathan (approved):
LGTM

deploy/olm-catalog/ibm-block-csi-operator/1.5.0/manifests/ibm-block-csi-operator.v1.5.0.clusterserviceversion.yaml
```diff
@@ -21,6 +21,9 @@ metadata:
             "name": "ibm-block-csi"
           },
           "spec": {
+            "license": {
+              "accept": "true",
+            },
```
oriyarde, line 26: is it only for red hat? or should it be added to community csv yaml as well?

matt-levan:
I think the idea is that the license defaults to false and the user must manually change it to true in a sort of acknowledging the acceptance of licenses used.

barakda:
A new PR will be opened after getting all the required information (like do we need to add license to community and build a license text/link for this purpose)

PR #147, opened 2021-03-24
CSI-2602 increase controller liveness time

oriyarde:
increase csi controller time to respond before restart, from 1 minute to 2.5 minutes.
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

### DIFF
--- a/pkg/controller/ibmblockcsi/syncer/csi_controller.go
+++ b/pkg/controller/ibmblockcsi/syncer/csi_controller.go
@@ -125,7 +125,7 @@ func (s *csiControllerSyncer) ensureContainersSpec() []corev1.Container {
 	})
 	controllerPlugin.ImagePullPolicy = s.driver.Spec.Controller.ImagePullPolicy
 
-	controllerPlugin.LivenessProbe = ensureProbe(10, 100, 2, corev1.Handler{
+	controllerPlugin.LivenessProbe = ensureProbe(10, 100, 5, corev1.Handler{
 		HTTPGet: &corev1.HTTPGetAction{
 			Path:   "/healthz",
 			Port:   controllerContainerHealthPort,


pkg/controller/ibmblockcsi/syncer/csi_controller.go
```diff
@@ -125,7 +125,7 @@ func (s *csiControllerSyncer) ensureContainersSpec() []corev1.Container {
 	})
 	controllerPlugin.ImagePullPolicy = s.driver.Spec.Controller.ImagePullPolicy
 
-	controllerPlugin.LivenessProbe = ensureProbe(10, 100, 2, corev1.Handler{
+	controllerPlugin.LivenessProbe = ensureProbe(10, 100, 5, corev1.Handler{
```
zingero, line 128: pkg/controller/ibmblockcsi/syncer/csi_controller.go:376:
```
func ensureProbe(delay, timeout, period int32, handler corev1.Handler) *corev1.Probe {
   return &corev1.Probe{
      InitialDelaySeconds: delay,
      TimeoutSeconds:      timeout,
      PeriodSeconds:       period,
      Handler:             handler,
      SuccessThreshold:    1,
      FailureThreshold:    30,
   }
}
```
The Probe struct says:
```
// How often (in seconds) to perform the probe.
// Default to 10 seconds. Minimum value is 1.
// +optional
PeriodSeconds int32 `json:"periodSeconds,omitempty" protobuf:"varint,4,opt,name=periodSeconds"`
```
It doesn't seem that 2/5 are 1/2.5 minute(s) accordingly as suggested in [this](https://github.com/IBM/ibm-block-csi-operator/pull/147#issue-599884450) comment.

oriyarde: why not? in the link it says `failureThreshold * periodSeconds`

zingero: Why don't we want to increase the `failureThreshold` value instead?
Increasing the `periodSeconds` value makes us wait longer than we should (up to 5 seconds instead of up to 2 seconds)

oriyarde: wait (longer than we should) for what?
how much should we wait, and why?

zingero: As far as I understand, we check a condition periodically in a loop.
Before this change, we checked every two seconds 30 times.
After this change, we check every five seconds 30 times.
The issue occurs because we don't wait enough time, right?
We can increase the checking interval or the checking amount.
For example, if we increase the interval from one second to one minute, we'll wait an average of 30 seconds before exiting this checking loop. Instead, we can increase the checking amount from 30 to 1800 and keep the interval (1 second), so the average time before exiting the checking loop would be half a second.

oriyarde: 1. if by "checking interval" you mean `periodSeconds`, then it's current value is 2 (not 1).
2. what is the downside of "wait an average of 30 seconds before exiting this checking loop"?
3. wouldn't your suggestion be more resource consuming? the default value is 10

zingero (approved):

ArbelNathan (approved):
LGTM

PR #153, opened 2021-05-04
Task/check seamless upgrade

zingero:


oriyarde (commented):
if this PR is open (and not a draft) - then is it ready for review?
if so, what about aligning the other CRD duplications?

zingero:
> if this PR is open (and not a draft) - then is it ready for review?
> if so, what about aligning the other CRD duplications?

This PR is irrelevant and I'll close it soon

PR #32, opened 2019-11-08
increase resource limits

27149chen:

PR #157, opened 2021-05-13
add community checks to operator

matancarmeli7:


.github/workflow/csi_operator_ci.yaml
```diff
@@ -0,0 +1,226 @@
+name: Operator test
+
+env:
+  OP_TEST_DEBUG: 1
+  OP_TEST_CONTAINER_OPT: "-t"
+on:
+  push:
+    pull_request:
+      - develop
```
oriyarde, line 9: should we add it to master as well?

build/ci/op-test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+OP_TEST_SCRIPT_URL=${OP_TEST_SCRIPT_URL-"https://raw.githubusercontent.com/operator-framework/operator-test-playbooks/master/upstream/test/test.sh"}
+
+bash <(curl -sL $OP_TEST_SCRIPT_URL) $*
+rc=$?
+echo "rc=$rc"
```
oriyarde, line 8: consider moving this to the next line, when rc==0

.github/workflow/csi_operator_ci.yaml
matancarmeli7: I thought about it, and before we will merge to master, the code will already be checked because there won't be changed in the code from the last PR to develop and from the PR to master.
But it doesn't cost much to add it for extra safety, so I can add it

oriyarde: thanks

.github/workflow/csi_operator_ci.yaml
```diff
@@ -0,0 +1,226 @@
...
+          build/ci/update_csv.sh
+        env:
+          current_operator_image: ${{ needs.prepare_env.outputs.current_operator_image }}
+          wanted_operator_image: ${{ needs.prepare_env.outputs.wanted_operator_image }}
+          csv_file: ${{ needs.prepare_env.outputs.csv_file }}
+          repository_path: ${{ needs.prepare_env.outputs.repository_path }}
+      - name: Operator test
+        run: |
+          echo "orange_v4.8 ${{ needs.prepare_env.outputs.community_operators_path }}"
+          build/ci/op-test.sh orange_v4.8 "${{ needs.prepare_env.outputs.community_operators_path }}"
```
oriyarde, line 226: looks like a new line at EOF is missing

build/ci/update_csv.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -xe
+set +o pipefail
+
+sed -i "s+$current_operator_image+$wanted_operator_image+g" $csv_file
+sed -i 's/.*minKubeVersion.*/  minKubeVersion: 1.19.0/' $csv_file
```
oriyarde, line 5: do we plan to maintain this from now on? and if not, what is our plan for the future?

.github/workflow/csi_operator_ci.yaml
```diff
@@ -0,0 +1,226 @@
...
+  CI-csi-kiwi:
+    name: "kiwi / Full operator test"
+    needs: prepare_env
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: update_csv
+        run: |
+          build/ci/update_csv.sh
```
oriyarde, line 57: could you open a ticket to reduce the code duplication, and write its number here?

build/ci/update_csv.sh
oriyarde: could we maybe add some skip logic if our `minKubeVersion` doesn't support a specific KinD version?

oriyarde (commented):
don't forget to add the ticket number in the PR title

build/ci/update_csv.sh
matancarmeli7: changing the minKubeVersion version is like putting a skip on the K8S version because it always will be in the run with a minimum K8S version of 1.19, but we will always use a newer one and it won't care what the version

.github/workflow/csi_operator_ci.yaml
matancarmeli7: done

build/ci/op-test.sh
matancarmeli7: done

.github/workflow/csi_operator_ci.yaml
matancarmeli7: done

.github/workflow/csi_operator_ci.yaml
matancarmeli7: done

build/ci/update_csv.sh
oriyarde: then I think a skip is more appropriate, since there is no special significance for 1.19

categories.json
```diff
@@ -0,0 +1,20 @@
...
+       "Logging & Tracing",
+       "Modernization & Migration",
+       "Monitoring",
+       "Networking",
+       "OpenShift Optional",
+       "Security",
+       "Storage",
+       "Streaming & Messaging"
+    ]
+ }
```
zingero, line 20: NL at EOF is missing

deploy/olm-catalog/ibm-block-csi-operator-community/ci.yaml
```diff
@@ -0,0 +1,3 @@
+---
+# Use `replaces-mode` or `semver-mode`. Once you switch to `semver-mode`, there is no easy way back.
+updateGraph: replaces-mode
```
zingero, line 3: NL at EOF is missing

zingero (commented):
almost :)

categories.json
```diff
@@ -0,0 +1,20 @@
+{
```
oriyarde, line 1: why is this file needed? does it have to be in the project root?
I'm starting to think we might need a separate project for this, or use a different approach, e.g. fork their repo

matancarmeli7: done

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,71 @@
+name: Community Operators test
+on:
+  push:
+    branches:
+      - task/CSI-2892_cert_community_operator_hub
```
oriyarde, line 5: is this for the target branch, private branch, or both?
what should be the value here before this PR merge?

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+jobs:
+  operator_image_build:
+   name: "set env"
+   runs-on: ubuntu-latest
+   steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+   outputs:
+     operator_image: "csiblock1/ibm-block-csi-operator-amd64:new"
```
oriyarde, line 14: for now I think we should use latest GA image, instead of this one.
could you open a ticket to build and push images in GitHub and write its number here?
(so this value would be dynamic in the future)

build/ci/clean_demo_pr.sh
```diff
@@ -0,0 +1,42 @@
...
+    echo "there are still checks that are running..."
+  done
+  if [ "$(gh pr checks $community_operators_openshift_branch --repo $github_csiblock_community_operators_repository | grep -i fail | wc -l)" -gt 0 ]
+  then
+    export failed_openshift_checks=true
+  fi
+  echo "The Openshift checks:"
+  gh pr checks $community_operators_openshift_branch --repo $github_csiblock_community_operators_repository
+  gh pr close $community_operators_openshift_branch --delete-branch --repo $github_csiblock_community_operators_repository
+fi
```
oriyarde, line 37: could you try to reduce this code dup into 1 function?

build/ci/create_demo_pr.sh
```diff
@@ -0,0 +1,44 @@
...
+
+sed -i "s+$current_operator_image+$wanted_operator_image+g" $csv_file
+
+# install gh command
+curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo gpg --dearmor -o /usr/share/keyrings/githubcli-archive-keyring.gpg
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+sudo apt update
+sudo apt install gh
+
+echo $github_token > mytoken.txt
```
oriyarde, line 12: `mytoken.txt` -> `github_token.txt`

build/ci/create_demo_pr.sh
```diff
@@ -0,0 +1,44 @@
...
+if [[ "$repo_pr" == *"$community_operators_openshift_branch"* ]]; then
+  gh pr close $community_operators_openshift_branch --delete-branch --repo $github_csiblock_community_operators_repository
+fi
+git checkout master
+git checkout -b $community_operators_openshift_branch
+yes | cp -r $repository_path/deploy/olm-catalog/ibm-block-csi-operator-community/ $repository_path/community-operators-fork/community-operators/
+git add .
+git commit --signoff -m "build number $github_build_number openshift"
+git push origin $community_operators_openshift_branch
+gh pr create --title "IBM Block CSI update to v1.6.0 operator" --repo $github_csiblock_community_operators_repository --base master --head $community_operators_openshift_branch --body "pr check"
```
oriyarde, line 44: I don't think the version is needed here

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - uses: FranzDiebold/github-env-vars-action@v2
+      - name: prepare_env
+        id: environment_setup
+        env:
+          operator_image: ${{ needs.operator_image_build.outputs.operator_image }}
+        run: |
+          build/ci/prepare_env.sh
+   outputs:
+     current_operator_image: "${{ steps.environment_setup.outputs.current_operator_image }}"
+     wanted_operator_image: "${{ steps.environment_setup.outputs.wanted_operator_image }}"
```
oriyarde, line 38: if I understand them correctly, consider renaming to something like `operator_image_in_branch` and `operator_image_for_test` respectively

build/ci/create_demo_pr.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -xe
+set +o pipefail
+
+sed -i "s+$current_operator_image+$wanted_operator_image+g" $csv_file
```
oriyarde, line 4: considered using a structure-based tool, e.g. `yq`

build/ci/prepare_env.sh
```diff
@@ -0,0 +1,20 @@
...
+export current_image=`echo ${current_image//\"}`
+export upstream_community_operators_path=upstream-community-operators/ibm-block-csi-operator-community/$csv_version
+export community_operators_path=community-operators/ibm-block-csi-operator-community/$csv_version
+
+echo "::set-output name=current_operator_image::${current_image}"
+echo "::set-output name=wanted_operator_image::${wanted_image}"
+echo "::set-output name=csv_file::${csv_file}"
+echo "::set-output name=upstream_community_operators_path::${upstream_community_operators_path}"
+echo "::set-output name=community_operators_path::${community_operators_path}"
+echo "::set-output name=repository_path::${repository_path}"
```
oriyarde, line 20: could we have a link to the actual tests?
e.g. https://github.com/csiblock/community-operators/pull/60/checks

build/ci/create_demo_pr.sh
oriyarde, line 44: could you reduce this into 1 function?

build/ci/clean_demo_pr.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash -xe
+set +o pipefail
+
+export failed_k8s_checks=false
+export failed_openshift_checks=false
```
oriyarde, line 5: consider using `passed` instead of `failed`, so that if `fail` (asserted in `grep -i fail`) ever changes, we won't have a false positive

build/ci/clean_demo_pr.sh
```diff
@@ -0,0 +1,42 @@
...
+    export failed_openshift_checks=true
+  fi
+  echo "The Openshift checks:"
+  gh pr checks $community_operators_openshift_branch --repo $github_csiblock_community_operators_repository
+  gh pr close $community_operators_openshift_branch --delete-branch --repo $github_csiblock_community_operators_repository
+fi
+
+if [ $failed_k8s_checks == "true" ] || [ $failed_openshift_checks == "true" ]
+then
+  exit 1
```
oriyarde, line 41: I'm not sure a `clean (up) pr ` script should fail because a test in it failed.
perhaps this file should be split into two different scripts

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+     csv_file: "${{ steps.environment_setup.outputs.csv_file }}"
+     community_operators_path: "${{ steps.environment_setup.outputs.community_operators_path }}"
+     upstream_community_operators_path: "${{ steps.environment_setup.outputs.upstream_community_operators_path }}"
+     repository_path: "${{ steps.environment_setup.outputs.repository_path }}"
+
+  CI-csi-demo-pr:
+    name: "create demo pr in forked community-operators repo"
```
oriyarde, line 45: consider reducing to something like `community-operators fork checks`

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+        env:
+          github_csiblock_community_operators_repository: "csiblock/community-operators"
+          community_operators_kubernetes_branch: "ibm-block-csi-update-kubernetes-${{ github.run_number }}"
+          community_operators_openshift_branch: "ibm-block-csi-update-openshift-${{ github.run_number }}"
+          github_token: ${{ secrets.CSIBLOCK_GITHUB_TOKEN }}
+          current_operator_image: ${{ needs.prepare_env.outputs.current_operator_image }}
+          wanted_operator_image: ${{ needs.prepare_env.outputs.wanted_operator_image }}
+          csv_file: ${{ needs.prepare_env.outputs.csv_file }}
+          repository_path: ${{ needs.prepare_env.outputs.repository_path }}
+      - name: clean demo pr
```
oriyarde, line 64: PR*

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+  CI-csi-demo-pr:
+    name: "create demo pr in forked community-operators repo"
+    needs: prepare_env
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: create demo PR
```
oriyarde, line 52: please align all job and step names to start with either an upper or lower case letter

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,71 @@
+name: Community Operators test
+on:
+  push:
+    branches:
+      - task/CSI-2892_cert_community_operator_hub
+jobs:
+  operator_image_build:
```
oriyarde, line 7: how come I can't find a run of this job at current checks?

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,71 @@
...
+   name: "set env"
+   runs-on: ubuntu-latest
+   steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+   outputs:
+     operator_image: "csiblock1/ibm-block-csi-operator-amd64:new"
+
+  prepare_env:
+   name: "prepare_env"
```
oriyarde, line 17: this sounds similar to "set env" (except for the underscore_convention), and it makes most of the jobs be **not** about the main checks, but "around" them, i.e. noise.
could we unify them?

.github/workflows/csi_operator_ci.yaml
```diff
@@ -1,13 +1,8 @@
-name: Operator test
-
-env:
-  OP_TEST_DEBUG: 1
-  OP_TEST_CONTAINER_OPT: "-t"
+name: Community Operators test
 on:
-  pull_request:
+  push:
```
zingero, line 3: Why do we want to do it on `push` instead of PR?

build/ci/clean_demo_pr.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash -xe
+set +o pipefail
+
+export failed_k8s_checks=false
+export failed_openshift_checks=false
+
+export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_kubernetes_branch`
+if [[ "$repo_pr" == *"$community_operators_kubernetes_branch"* ]]; then
+  sleep 5
```
zingero, line 9: Why do we need to sleep 5 seconds?

build/ci/create_demo_pr.sh
zingero: Also in #32

build/ci/create_demo_pr.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# install gh command
+curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo gpg --dearmor -o /usr/share/keyrings/githubcli-archive-keyring.gpg
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+sudo apt update
+sudo apt install gh
```
oriyarde, line 10: `WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`
consider using `apt-get`?

build/ci/create_demo_pr.sh
oriyarde: @zingero there is [already a conversation](https://github.com/IBM/ibm-block-csi-operator/pull/157#discussion_r648266872) for this

oriyarde: and in general, these line links inside diffs lead to the line, but the line is hidden by the filename, so I think regular line links are better, e.g. in this case [#32](https://github.com/IBM/ibm-block-csi-operator/blob/352247c243f8dd91047b0ba6a123387acd678902/build/ci/create_demo_pr.sh#L32)

.github/workflows/csi_operator_ci.yaml
oriyarde: why not both?

build/ci/create_demo_pr.sh
matancarmeli7: done

build/ci/clean_demo_pr.sh
matancarmeli7: done

build/ci/clean_demo_pr.sh
oriyarde: another script (in a separate step) that does all the `passed` checking.
this new script (and step) shall run before the clean up

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Prepare_env
+        id: environment_setup
+        run: |
+          build/ci/prepare_env.sh
+   outputs:
+     community_operators_path: "${{ steps.environment_setup.outputs.community_operators_path }}"
+     upstream_community_operators_path: "${{ steps.environment_setup.outputs.upstream_community_operators_path }}"
+     repository_path: "${{ steps.environment_setup.outputs.repository_path }}"
+     operator_image_in_branch: "${{ steps.environment_setup.outputs.operator_image_in_branch }}"
+     operator_image_for_test: ${{ needs.operator_image_build.outputs.operator_image }}
```
oriyarde, line 40: `operator_image` -> `operator_image_for_test`

build/ci/create_demo_pr.sh
oriyarde: did you use [`|=`](https://mikefarah.gitbook.io/yq/operators/assign-update#updated-multiple-paths)?
what is the error?

build/ci/prepare_env.sh
oriyarde: if any of their tests fail, where can I find a link to the details?

.github/workflows/csi_operator_ci.yaml
matancarmeli7: because the last code before the merge will be the same when it will be in `develop` so we will test the same thing twice

build/ci/prepare_env.sh
matancarmeli7: you will have a link in the checks
image

.github/workflows/csi_operator_ci.yaml
oriyarde: unless there are missing changes from develop in the private branch...

build/ci/create_demo_pr.sh
matancarmeli7: which error?

.github/workflows/csi_operator_ci.yaml
oriyarde: could we include all latest changes from develop in the beginning of the test?

matancarmeli7: you would have to merge develop into your private branch before pushing to develop, no?

oriyarde: it is not being enforced

build/ci/check_all_demo_pr_checks_were_passed.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash -xe
```
oriyarde, line 1: `check_all_demo_pr_checks_were_passed.sh` -> `check_demo_pr_checks.sh`

build/ci/create_demo_pr.sh
oriyarde: I assumed it didn't succeed because of an error.
could you write here what `yq` command should, but doesn't work?

build/ci/create_demo_pr.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash -xe
+set +o pipefail
+
+sed -i "s+$operator_image_in_branch+$operator_image_for_test+g" $csv_file
+
+# install gh command
+curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo gpg --dearmor -o /usr/share/keyrings/githubcli-archive-keyring.gpg
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+sudo apt-get update
+sudo apt-get install gh
```
oriyarde, line 10: shouldn't this be part of the prepare/setup env job (similiar to `yq`)?

.github/workflows/csi_operator_ci.yaml
oriyarde: ok, could you align an underscore_convention for their names?

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,31 @@
...
+wait_for_checks_to_complete(){
+  community_operators_branch=$1
+  all_checks_passed=true
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
+    while [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i pending | wc -l)" -gt 0 ]; do
+      sleep 5
+    done
+    if [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i fail | wc -l)" -gt 0 ]
```
oriyarde, line 16: could you grep for "not pass" instead of "fail"? to avoid potential false positives

.github/workflows/csi_operator_ci.yaml
oriyarde: could we enforce it?

build/ci/prepare_env.sh
oriyarde: ok, good enough

build/ci/create_demo_pr.sh
matancarmeli7: I don't remember the command that I used (it was two weeks ago) but I remember that the `yq` command didn't parse our YAML good in the alm-examples field

.github/workflows/csi_operator_ci.yaml
matancarmeli7: added the trigger on push to develop and master

build/ci/clean_demo_pr.sh
zingero: We sleep for 10 seconds the first time.
Does it make sense?

build/ci/clean_demo_pr.sh
```diff
@@ -1,42 +1,18 @@
 #!/bin/bash -xe
 set +o pipefail
 
-export failed_k8s_checks=false
-export failed_openshift_checks=false
+sleep 20
```
zingero, line 4: 20 seconds?!
This sounds like a lot

build/ci/create_demo_pr.sh
oriyarde: oh, then alm-examples value is a json, so it probably needs an additional `jq` command.
let me know if you need any help with that

build/ci/create_demo_pr.sh
oriyarde: can [this](https://stackoverflow.com/a/57958803) help?

build/ci/clean_demo_pr.sh
matancarmeli7: removed it

build/ci/clean_demo_pr.sh
matancarmeli7: I put 5 seconds at the beginning just for insurance that the checks are running

build/ci/check_demo_pr_checks.sh
oriyarde: I meant something like `grep -iv pass`

build/ci/create_demo_pr.sh
matancarmeli7: done

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,84 @@
...
+        env:
+          github_csiblock_community_operators_repository: "csiblock/community-operators"
+          community_operators_kubernetes_branch: "ibm-block-csi-update-kubernetes-${{ github.run_number }}"
+          community_operators_openshift_branch: "ibm-block-csi-update-openshift-${{ github.run_number }}"
+          github_token: ${{ secrets.CSIBLOCK_GITHUB_TOKEN }}
+          repository_path: ${{ needs.prepare_env.outputs.repository_path }}
+          operator_image_in_branch: ${{ needs.prepare_env.outputs.operator_image_in_branch }}
+          operator_image_for_test: ${{ needs.prepare_env.outputs.operator_image_for_test }}
+          csv_file: ${{ needs.prepare_env.outputs.csv_file }}
+      - name: Check if all the demo PR checks were passed
```
oriyarde, line 70: consider shortening to `Assert all demo PR checks passed`

matancarmeli7: done

build/ci/check_demo_pr_checks.sh
matancarmeli7: done

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,84 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - name: Install dependencies
+        run: python -m pip install --upgrade pip yq
+      - name: install apt dependencies
+        run: |
+          build/ci/install_ci_dependencies.sh
+      - uses: FranzDiebold/github-env-vars-action@v2
+      - name: Prepare_env
```
oriyarde, line 38: please remove the_underscore

matancarmeli7: done

build/ci/install_ci_dependencies.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -xe
+set +o pipefail
+
+# install gh command
+curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo gpg --dearmor -o /usr/share/keyrings/githubcli-archive-keyring.gpg
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+sudo apt-get update
+sudo apt-get install gh
```
oriyarde, line 8: this seems to be missing a newline at EOF

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,46 @@
...
+      sleep 15
+    done
+    summary_counter=0
+    export search_summary_test=`gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i summary`
+    while [ "$search_summary_test" != *"pending"* ] && [ $summary_counter -lt 20 ]; do
+      sleep 1
+      export search_summary_test=`gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i summary`
+      ((summary_counter=summary_counter+1))
+      if [[ "$search_summary_test" == *"pass"* ]]; then
+        summary_counter=20
```
oriyarde, line 23: what does `20` mean here?

build/ci/install_ci_dependencies.sh
matancarmeli7: done

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash -xe
+set +o pipefail
+
+export passed_k8s_checks=true
+export passed_openshift_checks=true
+
+wait_for_checks_to_complete(){
+  community_operators_branch=$1
+  all_checks_passed=true
```
oriyarde, line 9: I would start with `false`, just to be on the safe side

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,46 @@
...
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
+    while [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i pending | wc -l)" -gt 0 ]; do
+      sleep 15
+    done
+    summary_counter=0
+    export search_summary_test=`gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i summary`
```
oriyarde, line 17: can't this be initialized with a simpler, constant value?

build/ci/check_demo_pr_checks.sh
matancarmeli7: to break the while loop and "tell" the `if` at line 26 that the summary test has passed

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,46 @@
...
+      if [[ "$search_summary_test" == *"pass"* ]]; then
+        summary_counter=20
+      fi
+    done
+    if [[ $summary_counter -lt 20 ]]; then
+      while [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i pending | wc -l)" -gt 0 ]; do
+      sleep 5
+    done
+    fi
+    if [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -iv pass | wc -l)" -gt 0 ]
```
oriyarde, line 31: this shouldn't be under `wait_for_checks_to_complete`, consider moving it to a new function, e.g. `check_all_checks`

build/ci/check_demo_pr_checks.sh
oriyarde: I don't see how this is related to the number `20`.
sounds like a `is_summary_test_passed` boolean to me

build/ci/create_demo_pr.sh
```diff
@@ -0,0 +1,35 @@
...
+echo $github_token > github_token.txt
+gh auth login --with-token < github_token.txt
+gh repo fork operator-framework/community-operators --clone community-operators-fork
+cd community-operators-fork
+git remote set-url origin https://csiblock:$github_token@github.com/csiblock/community-operators.git
+git fetch upstream
+git rebase upstream/master
+git push origin master --force
+git config --global user.email csi.block1@il.ibm.com
+git config --global user.name csiblock
```
oriyarde, line 15: shouldn't this also be in prepare env?

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,84 @@
...
+        uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - name: Install dependencies
+        run: python -m pip install --upgrade pip yq
+      - name: install apt dependencies
+        run: |
+          build/ci/install_ci_dependencies.sh
```
oriyarde, line 36: `install` -> `Install`
why are these even separated?
if there's a good reason, then at least consider renaming the 1st to something more indicative, depending on the reason

matancarmeli7: done

build/ci/check_demo_pr_checks.sh
matancarmeli7: I don't think so, its the only way we can get the status of checks

build/ci/check_demo_pr_checks.sh
matancarmeli7: done

build/ci/check_demo_pr_checks.sh
matancarmeli7: done

build/ci/check_demo_pr_checks.sh
matancarmeli7: done

build/ci/create_demo_pr.sh
matancarmeli7: I can't pass it to another stage and it will stay for the next stage

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,54 @@
...
+    done
+    if [[ $is_summary_test_passed == "false" ]]; then
+      while [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i pending | wc -l)" -gt 0 ]; do
+      sleep 5
+    done
+    fi
+  fi
+}
+
+are_all_tests_were_passed(){
```
oriyarde, line 34: `are_all_tests_were_passed` -> `check_all_checks`/`did_all_checks_pass`

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,54 @@
...
+      sleep 1
+      export search_summary_test=`gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i summary`
+      ((summary_counter=summary_counter+1))
+      if [[ "$search_summary_test" == *"pass"* ]]; then
+        is_summary_test_passed=true
+      fi
+    done
+    if [[ $is_summary_test_passed == "false" ]]; then
+      while [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i pending | wc -l)" -gt 0 ]; do
+      sleep 5
```
oriyarde, line 28: how can this be, after the [same `while`](https://github.com/IBM/ibm-block-csi-operator/blob/f58e6f493eeba993456e6786e97eb3cbb6c08c78/build/ci/check_demo_pr_checks.sh#L13) already finished?

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,54 @@
...
+export passed_k8s_checks=true
+export passed_openshift_checks=true
+
+wait_for_checks_to_complete(){
+  community_operators_branch=$1
+  is_summary_test_passed=false
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
+    while [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i pending | wc -l)" -gt 0 ]; do
```
oriyarde, line 13: this long `gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository` command repeats a lot here, please put it in a var

build/ci/check_demo_pr_checks.sh
oriyarde: I meant that it's already being executed [inside the `while`](https://github.com/IBM/ibm-block-csi-operator/blob/f58e6f493eeba993456e6786e97eb3cbb6c08c78/build/ci/check_demo_pr_checks.sh#L20), so I don't think we need to execute it here

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,54 @@
...
+      sleep 15
+    done
+    summary_counter=0
+    export search_summary_test=`gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i summary`
+    while [ "$search_summary_test" != *"pending"* ] && [ $summary_counter -lt 20 ] && [ $is_summary_test_passed == "false" ]; do
+      sleep 1
+      export search_summary_test=`gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i summary`
```
oriyarde, line 20: `search_summary_test` -> `test_summary`

build/ci/check_demo_pr_checks.sh
```diff
@@ -0,0 +1,54 @@
...
+  is_summary_test_passed=false
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
+    while [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i pending | wc -l)" -gt 0 ]; do
+      sleep 15
+    done
+    summary_counter=0
+    export search_summary_test=`gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -i summary`
+    while [ "$search_summary_test" != *"pending"* ] && [ $summary_counter -lt 20 ] && [ $is_summary_test_passed == "false" ]; do
```
oriyarde, line 18: put `20` in an indicative const name

build/ci/check_demo_pr_checks.sh
oriyarde: could you move it to a different step as well?

build/ci/create_demo_pr.sh
```diff
@@ -0,0 +1,35 @@
...
+  cluster_kind=$3  
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    gh pr close $community_operators_branch --delete-branch --repo $github_csiblock_community_operators_repository
+  fi
+  git checkout master
+  git checkout -b $community_operators_branch
+  yes | cp -r $repository_path/deploy/olm-catalog/ibm-block-csi-operator-community/ $dest_path
+  git add .
+  git commit --signoff -m "build number $github_build_number $cluster_kind"
```
oriyarde, line 29: `$github_build_number` is not passed to this script

build/ci/create_demo_pr.sh
matancarmeli7: done
but I didn't managed to do it in a function because I couldn't pass variables to the field section

build/ci/check_demo_pr_checks.sh
matancarmeli7: I did, its under `Assert all demo PR checks passed` step

oriyarde: I meant could you make it before `Assert all demo PR checks passed`

build/ci/create_demo_pr.sh
oriyarde: please share the code that doesn't work (with error, if exists)

build/ci/create_demo_pr.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash -xe
+set +o pipefail
+
+yq -y --arg operator_image $operator_image_for_test '.spec.install.spec.deployments[0].spec.template.spec.containers[0].image |= $operator_image' $csv_file > temp.yaml
+cat temp.yaml > $csv_file
+yq -y --arg operator_image $operator_image_for_test '.metadata.annotations.containerImage |= $operator_image' $csv_file > temp.yaml
+cat temp.yaml > $csv_file
```
oriyarde, line 7: I don't think we need to cat every time

build/ci/setup_ci_dependencies.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# install gh command
+curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo gpg --dearmor -o /usr/share/keyrings/githubcli-archive-keyring.gpg
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+sudo apt-get update
+sudo apt-get install gh
+
+# configure git configuration
+git config --global user.email csi.block1@il.ibm.com
+git config --global user.name csiblock
```
oriyarde, line 15: new line is missing

build/ci/setup_ci_dependencies.sh
```diff
@@ -0,0 +1,15 @@
...
+python -m pip install --upgrade pip yq
+echo yq > dev-requirements.txt
+
+# install gh command
+curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo gpg --dearmor -o /usr/share/keyrings/githubcli-archive-keyring.gpg
+echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+sudo apt-get update
+sudo apt-get install gh
+
+# configure git configuration
```
oriyarde, line 13: no need for `documentation` (or a comment at all)

build/ci/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -xe
+set +o pipefail
+
+wait_for_checks_to_complete(){
+  community_operators_branch=$1
+  all_tests_passed=false
+  gh_pr_checks_command=(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository)
```
oriyarde, line 7: why is it an array and not simply a string?

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,99 @@
...
+          key: ${{ env.pythonLocation }}-${{ hashFiles('dev-requirements.txt') }}
+      - name: Retrieve gitconfig file
+        uses: actions/download-artifact@v2
+        with:
+          name: gitconfig-file
+          path: /home/runner
+      - name: Create demo PR
+        run: |
+          build/ci/create_demo_pr.sh
+      - name: Wait for all the checks to finish running
```
oriyarde, line 90: `Wait for all the checks to finish running` -> `Wait for all checks to finish`

build/ci/check_all_checks.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -xe
```
oriyarde, line 1: please move all new scripts to a new dir, e.g. `build/ci/community`

build/ci/create_demo_pr.sh
oriyarde: also `build number $github_build_number` -> `build #$github_build_number`

build/ci/prepare_env.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -xe
+set +o pipefail
+
+export csv_version=`yq .channels[0].currentCSV deploy/olm-catalog/ibm-block-csi-operator-community/ibm-block-csi-operator.package.yaml`
+export csv_version=`echo ${csv_version//ibm-block-csi-operator.v}`
+export csv_version=`echo ${csv_version//\"}`
```
oriyarde, line 6: do we really need to export in the 1st two lines?

build/ci/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
...
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
+    while [ `"${gh_pr_checks_command[@]}" | grep -i pending | wc -l` -gt 0 ]; do
+      sleep 15
+    done
+    summary_counter=0
+    seconds_to_wait_for_summary_test=20
+    while [ "$test_summary" != *"pending"* ] && [ $summary_counter -lt $seconds_to_wait_for_summary_test ] && [ $all_tests_passed == "false" ]; do
+      sleep 1
+      export test_summary=`"${gh_pr_checks_command[@]}" | grep -i summary`
```
oriyarde, line 18: do we really need to export here?

build/ci/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
...
+  all_tests_passed=false
+  gh_pr_checks_command=(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository)
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
+    while [ `"${gh_pr_checks_command[@]}" | grep -i pending | wc -l` -gt 0 ]; do
+      sleep 15
+    done
+    summary_counter=0
+    seconds_to_wait_for_summary_test=20
```
oriyarde, line 15: `summary_counter` -> `seconds_waited_for_summary_test`
and to avoid mixing them up:
`seconds_to_wait_for_summary_test` -> `summary_test_timeout_seconds`?

build/ci/check_demo_pr_checks.sh
oriyarde: there are 3 `while`s - I'm not referring to the second one.
why do we need this 3rd `while`?

matancarmeli7: I know, I meant to the third while

build/ci/create_demo_pr.sh
matancarmeli7: when I tried to put the input of the command directly to the `$csv_file`, it became empty

build/ci/create_demo_pr.sh
matancarmeli7: this was my last try
`yq --arg fieldtest $field -y '.[$fieldtest] |= "ni"' ibm-block-csi-operator.v1.6.0.clusterserviceversion.yaml > matan.yaml`

build/ci/community/check_all_checks.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -xe
+set +o pipefail
+
+export passed_k8s_checks=true
+export passed_openshift_checks=true
```
zingero, line 5:
```suggestion
export passed_k8s_checks=false
export passed_openshift_checks=false
```

build/ci/community/check_all_checks.sh
```diff
@@ -0,0 +1,24 @@
...
+
+did_all_checks_pass(){
+  community_operators_branch=$1
+  all_checks_passed=false
+  if [ "$(gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository | grep -iv pass | wc -l)" -eq 0 ]
```
zingero, line 10:
```suggestion
  if [ "$(gh pr checks $community_operators_branch --repo $forked_community_operators_repository | grep -iv pass | wc -l)" -eq 0 ]
```

build/ci/community/check_all_checks.sh
```diff
@@ -0,0 +1,24 @@
...
+  then
+    export all_checks_passed=true
```
zingero, line 12: Why do we export this value?

build/ci/community/clean_demo_pr.sh
```diff
@@ -1,15 +1,13 @@
...
-sleep 20
-
 print_checks_and_delete_pr(){
   community_operators_branch=$1
   cluster_kind=$2
   export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
   if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
     echo "The $cluster_kind checks:"
-    gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository
+    gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository || true
```
zingero, line 10: AFAIK the `set` command is here for that. Let's try to not abuse the `||` operator

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -xe
+set +o pipefail
+
+wait_for_checks_to_complete(){
+  community_operators_branch=$1
+  all_tests_passed=false
+  gh_pr_checks_command="gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository"
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
```
zingero, line 8: Do we need the `export`?

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
...
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
```
zingero, line 10: Let's try to reduce the sleeping time all over this function

build/ci/community/setup_ci_dependencies.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -xe
+set +o pipefail
+
+python -m pip install --upgrade pip yq
```
oriyarde, line 4: please use a specific version, in case they break backward compatibility

build/ci/create_demo_pr.sh
oriyarde: for latest `yq` version, this works for me (with `export` before, and without a `temp.yaml`):
`yq eval ".spec.relatedImages[0].image |= env(operator_image_for_test)" ibm-block-csi-operator.v1.6.0.clusterserviceversion.yaml -i`

build/ci/create_demo_pr.sh
oriyarde: I meant that simply one `cat` is sufficient (after the yaml file is finally ready)

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
...
+    while [ `eval $gh_pr_checks_command | grep -i pending | wc -l` -gt 0 ]; do
```
oriyarde, line 11: is `eval` really necessary?

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
...
+      sleep 15
+    done
+    seconds_waited_for_summary_test=0
+    seconds_to_wait_for_summary_test=20
+    while [ "$test_summary" != *"pending"* ] && [ $seconds_waited_for_summary_test -lt $summary_test_timeout_seconds] && [ $all_tests_passed == "false" ]; do
+      sleep 1
+      test_summary=`eval $gh_pr_checks_command | grep -i summary`
+      ((seconds_waited_for_summary_test=seconds_waited_for_summary_test+1))
```
oriyarde, line 19: the counter incrementation should come right after the sleep

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
...
+  gh_pr_checks_command="gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository"
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
+    while [ `eval $gh_pr_checks_command | grep -i pending | wc -l` -gt 0 ]; do
+      sleep 15
+    done
+    seconds_waited_for_summary_test=0
+    seconds_to_wait_for_summary_test=20
+    while [ "$test_summary" != *"pending"* ] && [ $seconds_waited_for_summary_test -lt $summary_test_timeout_seconds] && [ $all_tests_passed == "false" ]; do
```
oriyarde, line 16: why do I see both `seconds_to_wait_for_summary_test` and `summary_test_timeout_seconds`?

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
...
+  all_tests_passed=false
+  gh_pr_checks_command="gh pr checks $community_operators_branch --repo $github_csiblock_community_operators_repository"
+  export repo_pr=`gh pr list --repo $github_csiblock_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    sleep 5
+    while [ `eval $gh_pr_checks_command | grep -i pending | wc -l` -gt 0 ]; do
+      sleep 15
+    done
+    seconds_waited_for_summary_test=0
+    seconds_to_wait_for_summary_test=20
```
oriyarde, line 15: if I understand correctly, `summary_test` should be `test_summary` (everywhere)

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -xe
+set +o pipefail
+
+wait_for_checks_to_complete(){
```
zingero, line 4: Let's split this function to 3 smaller functions:

1. wait_for_tests_to_finish
2. collect_tests_results
3. wait_for_summary_to_finish

build/ci/wait_demo_pr_checks.sh
oriyarde: please remove all unneeded `export`s as well (if there are any)

build/ci/community/check_all_checks.sh
```diff
@@ -0,0 +1,24 @@
...
+  fi
+  echo "$all_checks_passed"
+}
+
+passed_k8s_checks=$(did_all_checks_pass $community_operators_kubernetes_branch)
+passed_openshift_checks=$(did_all_checks_pass $community_operators_openshift_branch)
```
oriyarde, line 18: why can't I see [these](https://github.com/operator-framework/community-operators/pull/4090#issuecomment-863226644) failures/runs?

build/ci/create_demo_pr.sh
matancarmeli7: done
but still didn't manage to fil the field section with parameter

build/ci/check_demo_pr_checks.sh
oriyarde: is it a summary test or a test summary?
since the `while` conditions are identical - what would happen if the 1st `while` will start after the summary already finished?

build/ci/create_demo_pr.sh
oriyarde:
```suggestion
declare -a operator_image_fields=(".spec.install.spec.deployments[0].spec.template.spec.containers[0].image" ".metadata.annotations.containerImage" ".spec.relatedImages[0].image")

for image_field in "${operator_image_fields[@]}"
do
    yq eval "$image_field |= env(operator_image_for_test)" $(basename $csv_file) -i
done
```

build/ci/wait_demo_pr_checks.sh
matancarmeli7: done

build/ci/community/setup_ci_dependencies.sh
oriyarde: pip is fine too, but `mikefarah/yq` still has implicit `:latest`.
the point is to avoid external breakage, regardless of which installation tool is used

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,26 @@
...
+  done
+}
+wait_for_checks_to_complete(){
+  community_operators_branch=$1
+  all_tests_passed=false
+  gh_pr_checks_command="gh pr checks $community_operators_branch --repo $forked_community_operators_repository"
+  repo_pr=`gh pr list --repo $forked_community_operators_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
+    wait_fot_checks_to_start $community_operators_branch $forked_community_operators_repository
+    while [[ ! "`$gh_pr_checks_command | grep -i summary`" =~ "pass" ]] && [[ ! "`$gh_pr_checks_command | grep -i summary`" =~ "fail" ]]; do
```
oriyarde, line 19: `$gh_pr_checks_command | grep -i summary` can be in a var

build/ci/community/prepare_env.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -xel
+set +o pipefail
+
+csv_version=`yq eval .channels[0].currentCSV deploy/olm-catalog/ibm-block-csi-operator-community/ibm-block-csi-operator.package.yaml`
```
oriyarde, line 4: `deploy/olm-catalog/ibm-block-csi-operator-community/ibm-block-csi-operator.package.yaml` can be in a var

build/ci/community/clean_demo_pr.sh
matancarmeli7: I want that only on this line the script won't exist because of a fail, because it just showing information, so I don't want to change the `set` that will change the behavior of the rest of the script

build/ci/community/check_all_checks.sh
oriyarde: from the logs it seems they are using [this](https://github.com/openshift/ci-tools).
could we try to run it somewhere as well? if not in cloud then at least in our clusters?

build/ci/community/prepare_env.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -xel
+set +o pipefail
+operator_package_file_path=deploy/olm-catalog/ibm-block-csi-operator-community/ibm-block-csi-operator.package.yaml
+csv_version=`yq eval .channels[0].currentCSV $operator_package_file_path`
```
oriyarde, line 4: will this test non-stable channels? e.g. 1.5.1

build/ci/wait_demo_pr_checks.sh
oriyarde: so why is [this one](https://github.com/IBM/ibm-block-csi-operator/blob/e52e5f95eba5ffc7666e47d9d6c5ea3dcdc0871b/build/ci/community/create_demo_pr.sh#L29) needed?

build/ci/community/wait_demo_pr_checks.sh
oriyarde: then maybe [wrap it in a function](https://unix.stackexchange.com/a/444952)?

build/ci/community/check_all_checks.sh
matancarmeli7: I think it should be in another ticket

build/ci/wait_demo_pr_checks.sh
matancarmeli7: oh I thought I removed it
done

build/ci/community/wait_demo_pr_checks.sh
matancarmeli7: done

build/ci/community/prepare_env.sh
matancarmeli7: no, we would want that it will be tested in the ci? because its a one time thing

build/ci/community/clean_demo_pr.sh
zingero: You can change it back

matancarmeli7: what do you mean?

build/ci/community/check_all_checks.sh
oriyarde: I think these missing tests is where we actually fail, so if we cover them in another ticket, that other ticket should be prioritized to be before this one.
if you think it is better, please open said ticket and write its number here, so we can resolve this

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,27 @@
...
+  forked_community_operators_repository=$2
+  gh_pr_checks_command="gh pr checks $community_operators_branch --repo $forked_community_operators_repository"
+  while [ `$gh_pr_checks_command | grep -i pending | wc -l` -eq 0 ]; do
+    sleep 1
+  done
+}
+wait_for_checks_to_complete(){
+  community_operators_branch=$1
+  all_tests_passed=false
+  gh_pr_checks_command="gh pr checks $community_operators_branch --repo $forked_community_operators_repository"
```
oriyarde, line 15: can `gh_pr_checks_command` be defined once for both functions?

build/ci/community/prepare_env.sh
oriyarde: I guess we need to find some way to test all non GA versions, but since versions like 1.5.1 are not common, I agree it's less important.
could you open a ticket for it and write its number here?

matancarmeli7: done
CSI-3171

build/ci/community/check_all_checks.sh
matancarmeli7: done
CSI-3172

build/ci/community/wait_demo_pr_checks.sh
matancarmeli7: done

build/ci/community/clean_demo_pr.sh
zingero: I suggest reading about the `-`/`+` flag prefixes

build/ci/wait_demo_pr_checks.sh
oriyarde: are they all necessary [here](https://github.com/IBM/ibm-block-csi-operator/blob/e4b57bfa9a6877b16ab85ecf561f6b34a6794e7b/build/ci/community/prepare_env.sh#L6-L10)?

build/ci/community/clean_demo_pr.sh
oriyarde: is it really better to set and unset a global state just for one command?

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash -xe
+set +o pipefail
+
+gh_pr_checks_command (){
+  community_operators_branch=$1
+  gh pr checks $community_operators_branch --repo $forked_community_operators_repository
+}
+
+wait_fot_checks_to_start(){
```
oriyarde, line 9: why is there no `for` in the filename?

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,30 @@
...
+  community_operators_branch=$1
+  forked_community_operators_repository=$2
```
oriyarde, line 11: is this necessary for `gh_pr_checks_command`?

build/ci/community/create_demo_pr.sh
```diff
@@ -0,0 +1,44 @@
...
+declare -a operator_image_fields=(
+    ".spec.install.spec.deployments[0].spec.template.spec.containers[0].image"
+    ".metadata.annotations.containerImage"
+    ".spec.relatedImages[0].image"
+)
+for image_field in "${operator_image_fields[@]}"
+do
+    yq eval "$image_field |= env(operator_image_for_test)" $(basename $csv_file) -i
+done
+cd -
```
oriyarde, line 15: haven't we had an "edit" function?

build/ci/community/create_demo_pr.sh
```diff
@@ -0,0 +1,44 @@
...
+
+echo $github_token > github_token.txt
+gh auth login --with-token < github_token.txt
+gh repo fork operator-framework/community-operators --clone community-operators-fork
+cd community-operators-fork
+git remote set-url origin https://csiblock:$github_token@github.com/csiblock/community-operators.git
+git fetch upstream
+git rebase upstream/master
+git push origin master --force
```
oriyarde, line 24: could this also be a function? with a name about whatever it doesn? something like fork_and_clone?

build/ci/community/check_all_checks.sh
```diff
@@ -1,21 +1,21 @@
 #!/bin/bash -xe
 set +o pipefail
 
-export passed_k8s_checks=true
-export passed_openshift_checks=true
+# CSI-3172 - run rad hat bot checks
```
zingero, line 4:
```suggestion
# CSI-3172 - run Red Hat bot checks
```

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -1,33 +1,32 @@
 #!/bin/bash -xe
 set +o pipefail
 
+gh_pr_checks_command (){
```
zingero, line 4: if we run the command, I suggest naming the function accordingly:
```suggestion
run_gh_pr_checks_command (){
```

build/ci/community/create_demo_pr.sh
```diff
@@ -1,38 +1,57 @@
...
   yes | cp -r $repository_path/deploy/olm-catalog/ibm-block-csi-operator-community/ $dest_path
   git add .
   git commit --signoff -m "build number $github_build_number $cluster_kind"
   git push origin $community_operators_branch
-  gh pr create --title "IBM Block CSI update $cluster_kind" --repo $github_csiblock_community_operators_repository --base master --head $community_operators_branch --body "pr check"
+  gh pr create --title "IBM Block CSI update $cluster_kind" --repo $forked_repository --base main --head $community_operators_branch --body "pr check"
+  cd -
+}
+
+update_community_operators_fork (){
```
zingero, line 39:
```suggestion
update_forked_community_operators_repo (){
```

zingero: @matancarmeli7 ?

build/ci/community/wait_demo_pr_checks.sh
oriyarde: isn't `gh_pr_checks` enough?

build/ci/community/create_demo_pr.sh
oriyarde: > you have a fork of the Spoon-Knife repository
from my understanding of english, `a fork of the X repo` == `repo X fork`

zingero: ok

build/ci/community/clean_demo_pr.sh
oriyarde: @zingero ?

build/ci/community/wait_demo_pr_checks.sh
oriyarde: the filename wasn't changed

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,32 @@
...
+}
+
+wait_for_checks_to_start(){
+  community_operators_branch=$1
+  forked_repository=$2
+  while [ `gh_pr_checks_command $community_operators_branch $forked_repository | wc -l` -eq 0 ]; do
+    sleep 1
+  done
+}
+wait_for_checks_to_complete(){
```
oriyarde, line 17: do we want to have a single line break between function? what is the convention here?
please try to use an IDE that helps with such conventions

build/ci/community/wait_demo_pr_checks.sh
oriyarde: I agree with the article, though I think in this case, the function is being used more like a command alias than a function.
I think this distinction is important, otherwise we might wrap all the commands we use with a long, low-level-detail named functions of the form `run_X_command`.
if you still disagree, at least consider the shorter `get_pr_checks`

build/ci/community/wait_demo_pr_checks.sh
oriyarde: if both are passed, then I think their name inside the `gh_pr_checks_command` function should be more general, e.g. `branch` and `repository`

build/ci/community/wait_demo_pr_checks.sh
```diff
@@ -0,0 +1,32 @@
...
+  community_operators_branch=$1
+  forked_repository=$2
+  all_tests_passed=false
+  repo_pr=`gh pr list --repo $forked_repository | grep $community_operators_branch`
+  if [[ "$repo_pr" == *"$community_operators_branch"* ]]; then
```
oriyarde, line 22: could you give an example for an output of the `gh pr list` command that would result in the `if` condition being false?

oriyarde: and what does it mean if there is no pr with the required branch? shouldn't it fail?

build/ci/community/check_all_checks.sh
oriyarde: `rad hat` would have been a much cooler name for them

build/ci/community/create_demo_pr.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -xel
+set +o pipefail
+
+edit_operator_image_in_csv_yaml_file (){
```
oriyarde, line 4: IMO no need for `_file`, since it's implied by the `yaml`

build/ci/community/create_demo_pr.sh
oriyarde: `update` seems pretty general to me.
would something like `sync_fork(_master_branch)?_from_upstream` be more accurate?

build/ci/community/setup_ci_dependencies.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -xe
+set +o pipefail
+
+echo $'yq() {\n  docker run --rm -e operator_image_for_test=$operator_image_for_test -i -v "${PWD}":/workdir mikefarah/yq:4 "$@"\n}' >> /home/runner/.bash_profile
```
oriyarde, line 4: please apply [conversation](https://github.com/IBM/ibm-block-csi-operator/pull/189#discussion_r680531659)s from other PRs where applicable

build/ci/community/wait_demo_pr_checks.sh
oriyarde: in addition, I see this code in at least 3 different places in this PR.
please move it to a common location

.github/workflows/csi_operator_ci.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+  CI-csi-demo-pr:
+    name: "community-operators fork checks"
+    needs: prepare_env
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      original_community_operators_repository: "k8s-operatorhub/community-operators"
+      original_community_operators_repository_prod: "redhat-openshift-ecosystem/community-operators-prod"
+      forked_community_operators_repository: "csiblock/community-operators"
```
oriyarde, line 59: please [change the `forked` terminology to `fork`](https://github.com/IBM/ibm-block-csi-operator/pull/157#discussion_r687627546) wherever applicable

build/ci/community/wait_demo_pr_checks.sh
zingero: `get_pr_checks` sounds great.
If you want to continue with the discussion about naming, I suggest having a session with Arbel.

build/ci/community/clean_demo_pr.sh
oriyarde: @zingero gentle reminder

build/ci/community/wait_demo_pr_checks.sh
oriyarde: @matancarmeli7 let's use `get_pr_checks` here

build/ci/community/clean_demo_pr.sh
zingero: Oh, sorry. I thought we're waiting for @matancarmeli7.
We can wrap only this command with the `set` command.
Which do you folks prefer?

oriyarde: ` || true` is shorter than two additional lines, but indeed hacky.
I think the current project convention is actually the shorter [`|| :`](https://github.com/IBM/ibm-block-csi-driver/blob/0f5ff540d6d8e69018201db2d00b34128fd3bae3/scripts/ci/build_push_images.sh#L27).
I prefer to stick to the convention, and if anyone wants, they can open a ticket to change the convention throughout the project

oriyarde (commented):
if this PR is not in plan to be merged in the near future, please change it to "draft"

PR #167, opened 2021-06-06
Bug/csi 3038 node pods in status creating due to bad secrets corelation

matancarmeli7:


pkg/internal/ibmblockcsi/ibmblockcsi.go
```diff
@@ -67,7 +68,7 @@ func (c *IBMBlockCSI) GetLabels() labels.Set {
 }
 
 // GetAnnotations returns all the annotations to be set on all resources
-func (c *IBMBlockCSI) GetAnnotations() labels.Set {
+func (c *IBMBlockCSI) GetAnnotations(ds_restarted_key string , ds_restarted_value string) labels.Set {
```
zingero, line 71: Functions should do one and only one thing.
In this case, the function should return the annotations.
Adding the key and value makes the function do two things.
Also, usually `get` functions don't change the data structure (`set functions do)

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -505,6 +525,58 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+		Namespace: instance.Namespace,
+	}, node)
+
+	if err != nil {
+		return err, node
+	}
+	return err, node
+}
+
+func (r *ReconcileIBMBlockCSI) getDriverPodsStatus(instance *ibmblockcsi.IBMBlockCSI,
```
zingero, line 564: What is the reason for returning two values?
I suggest breaking this function into two functions (one for the controller and one for the driver)
Then, the local variables will be redundant.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -505,6 +525,58 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+	node := &appsv1.DaemonSet{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{
+		Name:      oconfig.GetNameForResource(oconfig.CSINode, instance.Name),
+		Namespace: instance.Namespace,
+	}, node)
+
+	if err != nil {
+		return err, node
+	}
+	return err, node
```
zingero, line 561: Why do we need the `if` statements?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -505,6 +525,58 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+			controllerStatefulset := &appsv1.StatefulSet{}
+			err := r.client.Get(context.TODO(), types.NamespacedName{
+				Name:      oconfig.GetNameForResource(oconfig.CSIController, instance.Name),
+				Namespace: instance.Namespace,
+			}, controllerStatefulset)
+		
+			if err != nil {
+				return err, controllerStatefulset
+			}
+			return err, controllerStatefulset
```
zingero, line 548: Why do we need the if statements?

zingero: Also, usually, when we return the error and another value, the error is the second value.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
zingero: Also, usually, when we return the error and another value, the error is the second value.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -56,7 +56,14 @@ import (
...
 // ReconcileTime is the delay between reconciliations
-const ReconcileTime = 30 * time.Second
+const (
+	ReconcileTime = 30 * time.Second
+	Node = "node"
+	Controller = "controller"
+)
+
+var ds_restarted_key = ""
+var ds_restarted_value = ""
```
zingero, line 66: What is `ds`?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -320,60 +328,46 @@ func (r *ReconcileIBMBlockCSI) getAccessorAndFinalizerName(instance *ibmblockcsi
...
-	instance.Status.ControllerReady = controllerStatefulset.Status.ReadyReplicas == controllerStatefulset.Status.Replicas
-	instance.Status.NodeReady = node.Status.DesiredNumberScheduled == node.Status.NumberAvailable
+	if !instance.Status.ControllerReady {
+		err := r.getControllerPod(controllerStatefulset, controllerPod)
+		if err != nil {
+			logger.Error(err, "failed to get controller pod")
+			return err
+		}
+
+		for _, containerstatus := range controllerPod.Status.ContainerStatuses {
```
zingero, line 348: camelCase:
`containerStatus`

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -320,60 +328,46 @@ func (r *ReconcileIBMBlockCSI) getAccessorAndFinalizerName(instance *ibmblockcsi
...
+					logger.Info("controller requires restart",
+											"ReadyReplicas", controllerStatefulset.Status.ReadyReplicas,
+											"Replicas", controllerStatefulset.Status.Replicas)
+					r.restartControllerPod(controllerPod)
+				}
+			}
+		}
+	}
+
+	instance.Status.ControllerReady, instance.Status.NodeReady = r.getDriverPodsStatus(instance,
```
zingero, line 360: Why do we need to assign the values in the `instance.Status`?
The function already does that.
I see two options:
1. remove the assignment in the function
2. remove the assignment here and rename the function since it doesn't get anything.

zingero (changes requested):
Run the seamless upgrade job, please.

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: daemonSet

zingero: ohhh
I thought it's "data structure".
Please rename it accordingly.

pkg/internal/ibmblockcsi/ibmblockcsi.go
zingero: I know it returns one thing.
It does more than one thing though.
For more information:
https://stackoverflow.com/questions/475675/when-is-a-function-too-long

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -535,46 +534,38 @@ func (r *ReconcileIBMBlockCSI) getRestartedAtAnnotation(Annotations map[string]s
...
+	return node, err
 }
 
-func (r *ReconcileIBMBlockCSI) getDriverPodsStatus(instance *ibmblockcsi.IBMBlockCSI, 
-	controller *appsv1.StatefulSet, node *appsv1.DaemonSet) (bool, bool) {
+func (r *ReconcileIBMBlockCSI) getControllerPodsStatus(instance *ibmblockcsi.IBMBlockCSI, controller *appsv1.StatefulSet) bool {
 	ControllerReady := false
+
+	ControllerReady = controller.Status.ReadyReplicas == controller.Status.Replicas
+	return ControllerReady
```
zingero: Why do we need the local variable? 
```suggestion
	return controller.Status.ReadyReplicas == controller.Status.Replicas 
```

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -535,46 +534,38 @@ func (r *ReconcileIBMBlockCSI) getRestartedAtAnnotation(Annotations map[string]s
...
+}
+
+func (r *ReconcileIBMBlockCSI) getNodePodsStatus(instance *ibmblockcsi.IBMBlockCSI, node *appsv1.DaemonSet) bool {
 	NodeReady := false
```
zingero, line 565: Same

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
zingero: Also, the function name says that it returns the status, but actually, we return something else.
Rename (also for the node function)

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -355,25 +350,23 @@ func (r *ReconcileIBMBlockCSI) updateStatus(instance *ibmblockcsi.IBMBlockCSI, o
...
 			}
-			if originalStatus.ControllerReady || !r.areAllPodImagesSynced(controllerStatefulset, controllerPod) {
-				logger.Info("controller requires restart",
-							"ReadyReplicas", controllerStatefulset.Status.ReadyReplicas,
-							"Replicas", controllerStatefulset.Status.Replicas)
-				controllerRestart = true
+	
+			for _, containerStatus := range controllerPod.Status.ContainerStatuses {
+				if containerStatus.State.Waiting != nil {
+					if containerStatus.State.Waiting.Reason == "ImagePullBackOff" {
```
oriyarde, line 356: why is this needed?
I don't think we should restart the pod for each faulty container

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -505,6 +524,44 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+	return node, err
+}
+
+func (r *ReconcileIBMBlockCSI) doesControllerPodsRunning(instance *ibmblockcsi.IBMBlockCSI, controller *appsv1.StatefulSet) bool {
+	return controller.Status.ReadyReplicas == controller.Status.Replicas
+}
+
+func (r *ReconcileIBMBlockCSI) doesNodePodsRunning(instance *ibmblockcsi.IBMBlockCSI, node *appsv1.DaemonSet) bool {
+	return node.Status.DesiredNumberScheduled == node.Status.NumberAvailable
+}
```
oriyarde, line 563: why do they need the `instance`?
boolean functions usually start with `is`, and in this case the high-level question is `is<Component>Ready`

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -505,6 +524,44 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+
+func (r *ReconcileIBMBlockCSI) getNodeK8sObject(instance *ibmblockcsi.IBMBlockCSI) (*appsv1.DaemonSet, error) {
+	node := &appsv1.DaemonSet{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{
+		Name:      oconfig.GetNameForResource(oconfig.CSINode, instance.Name),
+		Namespace: instance.Namespace,
+	}, node)
+
+	return node, err
+}
```
oriyarde, line 555: rename `k8sObject` to the specific k8s object, and fix over indentation

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: we still need to handle the problem with statefulsets that doesn't redoply pods in ImagePullBackOff even though the user updates their images

oriyarde: oh, then you can still use `areAllPodImagesSynced` for that

matancarmeli7: what is `areAllPodImagesSynced`? this is an existing function? or its a function that you want me to add?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/internal/ibmblockcsi/ibmblockcsi.go
zingero: The function changes the DS. According to its name, it should be read-only (const function in C++)

matancarmeli7: I don't think that it changes the DS. Its only returns the wanted annotations that should be on the DS/STS

zingero: I meant DS == Data Structure, no Daemon Set.
Sorry for the confusion.

matancarmeli7: done

zingero (approved):

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: I still don't see why we should restart the pod for each faulty container, not why we need to check the waiting reason.
can't we just keep the original code here (minus the `originalStatus` check)?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: no need for the word `Object`

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: could you please renaming the functions to `isControllerRead` and `isNodeReady` respectively?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -355,25 +350,23 @@ func (r *ReconcileIBMBlockCSI) updateStatus(instance *ibmblockcsi.IBMBlockCSI, o
...
-			logger.Info("node rollout requires restart",
-				"DesiredNumberScheduled", node.Status.DesiredNumberScheduled,
-				"NumberAvailable", node.Status.NumberAvailable)
-			nodeRolloutRestart = true
-		}
 		phase = csiv1.DriverPhaseCreating
 	}
 	instance.Status.Phase = phase
 	instance.Status.Version = oversion.DriverVersion
+	logger.Info("updating IBMBlockCSI status")
```
oriyarde, line 369: I don't think this is true (it's true only inside the following `if`)

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: We wanted to remove the original code and move it to the service account reconciler.
The original code was also restarting the controller pods for each faulty container, this code is restarting only when there is an ImagePullBackOff and the function `areAllPodImagesSynced` returns false so it minimizes the restating action to less events

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -493,6 +468,50 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+				
+				if rErr != nil {
+					return rErr
+				}
+			}
+			if strings.Contains(sa.Name, Node) {
+				nodelogger := log.WithValues("Resource Type", "Node DaemonSet")
+				nodelogger.Info("node rollout requires restart",
+				"DesiredNumberScheduled", nodeDaemonSet.Status.DesiredNumberScheduled,				
+				"NumberAvailable", nodeDaemonSet.Status.NumberAvailable)
```
oriyarde, line 505: please indent function parameters wherever applicable, to align to current project convention

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: the original code did **not** restart the controller pod **for each** faulty container, it restarted the pod **once** if **any** container image was not in sync with the statefulset.
so current diff can potentially **add** restart events

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: do you have an example?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -493,6 +468,50 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+			if err != nil {
+				return err
+			}
+
+			nodeDaemonSet, err := r.getNodeDaemonSetObject(instance)
+			if err != nil {
+				return err
+			}
+
+			if strings.Contains(sa.Name, Controller) {
```
oriyarde, line 481: please compare between full strings (i.e. without `Contains`).
perhaps with [`GetNameForResource`](https://github.com/IBM/ibm-block-csi-operator/blob/9a30d70922dbfa46429c8ce6df9c16684510b76d/pkg/controller/ibmblockcsi/syncer/csi_controller.go#L109)

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: but the current diff also restart the pod once if any container image was not in sync with the statefulset, but when there is `ImagePullBackOff`, I can remove the if of the `ImagePullBackOff` if you like

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: e.g. [here](https://github.com/IBM/ibm-block-csi-operator/blob/9a30d70922dbfa46429c8ce6df9c16684510b76d/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go#L420) (though it should be even a bit more indented, to be aligned with the 1st argument)

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: it seems to me as if the current diff restarts **for each** faulty container.
what am I missing here?

oriyarde: I would remove the `ImagePullBackOff` regardless

matancarmeli7: I will do it like you suggested at the beginning

matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -56,7 +56,14 @@ import (
...
 
 // ReconcileTime is the delay between reconciliations
-const ReconcileTime = 30 * time.Second
+const (
+	ReconcileTime = 30 * time.Second
+	Node = "node"
+	Controller = "controller"
+)
+
+var daemonSet_restarted_key = ""
```
oriyarde, line 65: this can be a `fmt.Sprintf("%s/restartedAt", oconfig.APIGroup)` "const" here

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -56,7 +56,14 @@ import (
...
+var daemonSet_restarted_value = ""
```
oriyarde, line 66: use camelCase everywhere

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -56,7 +56,12 @@ import (
 )
 
 // ReconcileTime is the delay between reconciliations
-const ReconcileTime = 30 * time.Second
+const (
+	ReconcileTime = 30 * time.Second
+)
```
oriyarde, line 61: no need for round brackets (parenthesis)

matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -493,6 +460,50 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+			if err != nil {
+				return err
+			}
+
+			nodeDaemonSet, err := r.getNodeDaemonSet(instance)
+			if err != nil {
+				return err
+			}
+
+			if config.GetNameForResource(config.CSIControllerServiceAccount, instance.Name) == sa.Name {
```
oriyarde, line 474: please give this return value a name, before comparing it.
e.g. `controllerServiceAccountName` (same for node)

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: I didn't do it because I wanted that as long as the `restartedAt` annotation won't be on the `daemonset` it will stay empty, that the sycer will know that there isn't a new annotation that it needs to take care of

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: it can check the value, I don't think the key is interesting

oriyarde: oh, you don't want `reconcileServiceAccount` to return anything...

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -505,6 +517,44 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+}
+
+func (r *ReconcileIBMBlockCSI) getControllerStatefulSet(instance *ibmblockcsi.IBMBlockCSI) (*appsv1.StatefulSet, error) {
+			controllerStatefulset := &appsv1.StatefulSet{}
+			err := r.client.Get(context.TODO(), types.NamespacedName{
+				Name:      oconfig.GetNameForResource(oconfig.CSIController, instance.Name),
+				Namespace: instance.Namespace,
+			}, controllerStatefulset)
+
+			return controllerStatefulset, err
```
oriyarde, line 537: please align all indentations to current project conventions.
e.g. this seems over-indented to me.
I think Ctrl+Alt+L

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: `reconcileServiceAccount` return only error, because `daemonSetRestartedKey` and `daemonSetRestartedValue` are global variables

matancarmeli7: so I can check the value and move the to the const section if you like

oriyarde: I will try to find a better way

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -493,6 +461,50 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+				if rErr != nil {
+					return rErr
+				}
+			}
+			if nodeServiceAccountName == sa.Name {
+				nodelogger := log.WithValues("Resource Type", "Node DaemonSet")
+				nodelogger.Info("node rollout requires restart",
+								"DesiredNumberScheduled", nodeDaemonSet.Status.DesiredNumberScheduled,				
+								"NumberAvailable", nodeDaemonSet.Status.NumberAvailable)
+				nodelogger.Info("csi node stopped being ready - restarting it")
```
oriyarde, line 499: why the separate loggers? consider using this function's common logger

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -493,6 +461,50 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+			nodeDaemonSet, err := r.getNodeDaemonSet(instance)
+			if err != nil {
+				return err
+			}
+
+			if controllerServiceAccountName == sa.Name {
+				controllerlogger := log.WithValues("Resource Type", "Controller")
+				controllerPod := &corev1.Pod{}
+				err := r.getControllerPod(controllerStatefulset, controllerPod)
+				if err != nil {
```
oriyarde, line 479: what about `errors.IsNotFound(err)`? what if the driver just wasn't deployed yet?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: because the resource is different and in the logs I don't see `node rollout requires restart, resource service account` for example

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: could you open a ticket for it and write its number here?
I guess it won't be for this release

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -493,6 +464,52 @@ func (r *ReconcileIBMBlockCSI) reconcileServiceAccount(instance *ibmblockcsi.IBM
...
+			if err != nil {
+				return err
+			}
+
+			if controllerServiceAccountName == sa.Name {
+				controllerlogger := log.WithValues("Resource Type", "Controller")
+				controllerPod := &corev1.Pod{}
+				err := r.getControllerPod(controllerStatefulset, controllerPod)
+				if errors.IsNotFound(err) {
+					return nil
```
oriyarde, line 483: will this prevent from the second service account to be reconciled?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: I think maybe I will add the `if errors.IsNotFound(err)` in the `getControllerPod`, and remove it from checking here and in the other place that we call to `getControllerPod` function. What do you think?

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
oriyarde: it would still require every caller to check whether the pod is not nil, but I guess it's better than nothing?
I think here specifically the caller doesn't need the pod, so it would be better if it could call a function that doesn't require the pod as a parameter

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -58,6 +58,10 @@ import (
 // ReconcileTime is the delay between reconciliations
 const ReconcileTime = 30 * time.Second
 
+// ticket to redundant those vars - CSI-3071
```
oriyarde, line 61: to remove*

matancarmeli7: done

oriyarde: I don't see it

matancarmeli7: now it should be applied, I didn't write the done as part of the review

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: done

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
```diff
@@ -424,7 +388,36 @@ func (r *ReconcileIBMBlockCSI) areAllPodImagesSynced(controllerStatefulset *apps
...
+		return nil
+	} else if err != nil {
+		logger.Error(err, "failed to get controller pod")
+		return err
+	}
+
+	return r.restartControllerPodfromStatefulSet(logger, instance, controllerStatefulset, controllerPod)
+}
+
+func (r *ReconcileIBMBlockCSI) restartControllerPodfromStatefulSet(logger *logf.DelegatingLogger, instance *ibmblockcsi.IBMBlockCSI,
```
oriyarde, line 414: does it really need the `instance`?

oriyarde (approved):
nice job!

pkg/controller/ibmblockcsi/ibmblockcsi_controller.go
matancarmeli7: no, I will remove it

matancarmeli7: done

PR #192, opened 2021-07-11
Feature/csi 3085 topology enable feature gate

ArbelNathan:


cmd/manager/main.go
```diff
@@ -20,8 +20,13 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/IBM/ibm-block-csi-operator/pkg/controller/ibmblockcsi/syncer"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"os"
 	"runtime"
+	"strings"
```
zingero, line 29: Is there a code style rule for packages order?

cmd/manager/main.go
```diff
@@ -54,6 +59,8 @@ var (
 	metricsHost               = "0.0.0.0"
 	metricsPort         int32 = 8383
 	operatorMetricsPort int32 = 8686
+	topologyPrefixes          = [...]string{"topology.kubernetes.io", "topology.block.csi.ibm.com"}
+	topologyEnabled     bool
```
zingero, line 63: Why don't we set an initial value in contrast to other variables?

cmd/manager/main.go
```diff
@@ -204,3 +218,30 @@ func serveCRMetrics(cfg *rest.Config) error {
 	return nil
 }
 */
+func IsTopologyInUse(ctx context.Context) (bool, error) {
+	kubeConfig, err := rest.InClusterConfig()
+	if err != nil {
+		log.Info("unable to load in-cluster configuration: %v", err)
+		log.Info("skipping topology retrieval. we might not be in a k8s cluster")
```
zingero, line 225: Why the first log's severity is `info`? Shouldn't it be higher?
Can we merge these messages together?

pkg/controller/ibmblockcsi/syncer/csi_controller.go
```diff
@@ -50,6 +50,7 @@ const (
 )
 
 var controllerContainerHealthPort = intstr.FromInt(controllerContainerHealthPortNumber)
+var TopologyEnabled bool
```
zingero, line 53: Same

pkg/controller/ibmblockcsi/syncer/csi_controller.go
```diff
@@ -134,10 +135,14 @@ func (s *csiControllerSyncer) ensureContainersSpec() []corev1.Container {
 	})
 
 	// csi provisioner sidecar
+	provisionerArgs := []string{"--csi-address=$(ADDRESS)", "--v=5", "--timeout=30s", "--default-fstype=ext4"}
+	if TopologyEnabled {
+		provisionerArgs = append(provisionerArgs, "--feature-gates=Topology=true")
+	}
 	provisioner := s.ensureContainer(provisionerContainerName,
 		s.getCSIProvisionerImage(),
 		// TODO: make timeout configurable
```
zingero, line 144: Do we need this comment here?

zingero (changes requested):

zingero:
Where are the tests?

cmd/manager/main.go
ArbelNathan: the auto one (ctrl alt o) put it there..

zingero: @oriyarde Is it the best way at the moment to order the packages?

cmd/manager/main.go
ArbelNathan: cause it will have no meaning, its initiated before use

pkg/controller/ibmblockcsi/syncer/csi_controller.go
ArbelNathan: it already initiated in that stage

pkg/controller/ibmblockcsi/syncer/csi_controller.go
ArbelNathan: I dont know, I was not the one that put it

cmd/manager/main.go
ArbelNathan: it will log with `log.Error` in the parent method.
do you think it's problematic as two separated logs?

pkg/controller/ibmblockcsi/syncer/csi_controller.go
zingero: You moved the reference. I think we should remove it or move it to the relevant line

pkg/controller/ibmblockcsi/syncer/csi_controller.go
zingero: What do you mean "at the stage"? What happens in Go if we access an uninitialized variable? Is there a default value?

cmd/manager/main.go
zingero: Now, yes, it's initiated before we use it. Maintainance-wise, we need to remember there is an uninitialized variable here and IMO we should avoid remembering things like that.

cmd/manager/main.go
zingero: Why do we need three messages for one error?
Don't we prefer one message (maybe longer)?

cmd/manager/main.go
oriyarde: currently yes

cmd/manager/main.go
oriyarde: I don't think the operator can function if it can't get the kube config.
this should be more severe than "info", and shouldn't assume `we might not be in a k8s cluster` (in contrast to the driver), because we don't plan to ever support such scenario for this operator

cmd/manager/main.go
```diff
@@ -204,3 +218,30 @@ func serveCRMetrics(cfg *rest.Config) error {
 	return nil
 }
 */
+func IsTopologyInUse(ctx context.Context) (bool, error) {
+	kubeConfig, err := rest.InClusterConfig()
```
oriyarde, line 222: all these "client init" lines seem like a duplication of those in [getServerVersion](https://github.com/IBM/ibm-block-csi-operator/blob/7f0f44b70ffb6d72d98c6fcb5fa3734756f0e08f/pkg/controller/ibmblockcsi/ibmblockcsi_controller.go#L76) - could you try to unify them, and make them run only once?

pkg/controller/ibmblockcsi/syncer/csi_controller.go
oriyarde: I would move it to the relevant line for now, because it exists in several places, which are not relevant for this ticket

pkg/controller/ibmblockcsi/syncer/csi_controller.go
ArbelNathan: its initiated in [main](https://github.com/IBM/ibm-block-csi-operator/blob/c9f1d2b9256a44368432c7beafe5c30fabb8aa79/cmd/manager/main.go#L140)

cmd/manager/main.go
ArbelNathan: "uninitialized variable"?

zingero: an uninitialized variable is a declared variable without a value.

pkg/controller/ibmblockcsi/syncer/csi_controller.go
zingero: As I [said](https://github.com/IBM/ibm-block-csi-operator/pull/192#discussion_r667518312) previously, We don't need to remember these things.

cmd/manager/main.go
ArbelNathan: oh, I read unutilized..

ArbelNathan: I removed it from being a public var completely

cmd/manager/main.go
ArbelNathan: ok, true
removed logs there.

pkg/controller/ibmblockcsi/syncer/csi_controller.go
ArbelNathan: done

pkg/controller/ibmblockcsi/syncer/csi_controller.go
ArbelNathan: done

zingero (approved):

cmd/manager/main.go
ArbelNathan: done

oriyarde (approved):
Great Job!

pkg/util/kubernetes/kube_util.go
```diff
@@ -18,17 +18,37 @@ package kubernetes
...
+	"os"
 	"runtime"
 
 	"github.com/pkg/errors"
-
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
```
zingero, line 30: `logf`? why not `log`?

zingero: I prefer not using "util" files. 
I suggest moving the code to a different file (`kube_client.go`?)

ArbelNathan: this file already in use by ibmblockcsi_controller.go and it using a similar library so it seems right to be there.. (if you still think I should move it it's supposed to be a problem)
logf - its the convention in the code (more than 12 places name it like that)

zingero: 1. "util" files can be easily abused since we can define lots of things at "utilities". Thus, I prefer to reduce our "util" usage.
2. what is the meaning of `f` in `logf`?

ArbelNathan: 1. will move.
2. nor sure but it's in the [doc too](https://sdk.operatorframework.io/docs/building-operators/golang/references/logging/#a-simple-example)

ArbelNathan: done

zingero (approved):

oriyarde (approved):
LGTM

PR #207, opened 2021-09-12
Improvment/csi 3382 topology remove general topology key

ArbelNathan:


main.go
```diff
@@ -44,7 +44,7 @@ var (
 	scheme               = runtime.NewScheme()
 	setupLog             = ctrl.Log.WithName("setup")
 	watchNamespaceEnvVar = "WATCH_NAMESPACE"
-	topologyPrefixes     = [...]string{"topology.kubernetes.io", "topology.block.csi.ibm.com"}
+	topologyPrefix       = "topology.block.csi.ibm.com"
```
oriyarde, line 47: it can still be a list, with one item

ArbelNathan: for what purpose?

oriyarde: so it would be easy to add/remove prefixes

ArbelNathan: do we have any intention of doing it?

oriyarde: theoretically in the future (and definitely in this PR), sure

ArbelNathan: ok, done

oriyarde (approved):
LGTM

PR #226, opened 2021-12-01
Task/csi 3503 add verification scrips only

matancarmeli7:
Signed-off-by: matancarmeli7 <matan.carmeli7@gmail.com>


hack/get_information_helper.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash -e
+
```
oriyarde, line 2: `get` makes it sound like an action will be performed when executing this file.
how about `project_info.sh`?

matancarmeli7: done

hack/check-files-alignment.sh
```diff
@@ -0,0 +1,72 @@
...
+  diff -qr --exclude=bin $project_dirname $project_dirname-expected/
+  rm -rf $project_dirname-expected/
+  cd $project_dirname
+}
+
+verify_full_operator_yaml_is_aligned(){
+  echo "check full operator yaml alignment"
+  declare -A yaml_kinds_by_origin_yaml_path=(
+      ["config/rbac/role.yaml"]="ClusterRole"
+      ["config/crd/bases/csi.ibm.com_ibmblockcsis.yaml"]="CustomResourceDefinition"
```
oriyarde, line 40:
```suggestion
      [roles_yaml_path]="ClusterRole"
      [origin_crd_yaml_path]="CustomResourceDefinition"
```

hack/check-files-alignment.sh
```diff
@@ -0,0 +1,72 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+source hack/project_info.sh
+roles_yaml_path=config/rbac/role.yaml
+origin_crd_yaml_path=config/crd/bases/csi.ibm.com_ibmblockcsis.yaml
```
oriyarde, line 21:
```suggestion
roles_yaml_path=config/rbac/role.yaml
crd_yaml_path=config/crd/bases/csi.ibm.com_ibmblockcsis.yaml
```
(similar to why `roles_yaml_path` shouldn't be `origin_roles_yaml_path`)

matancarmeli7: I added the origin because there are other crd YAML files and there isn't any other role YAML file, so this is why I added the origin.
Do you still want me to change it? or keep it?

hack/check-files-alignment.sh
matancarmeli7: done

hack/check-files-alignment.sh
```diff
@@ -0,0 +1,72 @@
...
+  echo "check generation"
+  project_dirname=ibm-block-csi-operator
+  cd ..
+  cp -r $project_dirname ./$project_dirname-expected
+  cd $project_dirname-expected/
+  make update
+  cd ..
+  diff -qr --exclude=bin $project_dirname $project_dirname-expected/
+  rm -rf $project_dirname-expected/
+  cd $project_dirname
```
ArbelNathan, line 33: what the last cd is for?

hack/check-files-alignment.sh
oriyarde: I think "origin" can be useful only in a context dealing with destination (usually a multidoc yaml, e.g. a csv yaml).
so please change it here, but feel free to use "origin" in later code where there is destination as well, e.g.
```
verify_yaml_doc_in_multidoc_yaml (){
origin_crd_yaml_path = $1
destination_yaml_path = $2
.
.
}
verify_yaml_doc_in_multidoc_yaml crd_yaml_path operator_yaml_path
```
though IMO when it comes to the diff verifications in this file, they are more of an "inclusion" (/"containment") check.
I think source and destination are more suitable where an actual copy operation is performed, because there is more focus on the direction

hack/check-files-alignment.sh
matancarmeli7: to get back to the main operator directory

hack/check-files-alignment.sh
matancarmeli7: done, changed the `origin_crd_yaml_path` variable

hack/project_info.sh
```diff
@@ -0,0 +1,41 @@
...
+  fi
+}
+
+get_csv_files (){
+  current_csi_version=$(get_current_csi_version)
+  ls deploy/olm-catalog/*/$current_csi_version/manifests/ibm-block-csi-operator.v$current_csi_version.clusterserviceversion.yaml
+}
+
+get_bundle_crds (){
+  current_csi_version=$(get_current_csi_version)
```
oriyarde, line 39: code dup

hack/project_info.sh
```diff
@@ -0,0 +1,41 @@
...
+#
+
+operator_yaml_path=deploy/installer/generated/ibm-block-csi-operator.yaml
+
+get_current_csi_version (){
+  current_csi_version=$(cat version/version.go | grep -i driverversion | awk -F = '{print $2}')
+  echo ${current_csi_version//\"}
+}
+
+are_manifest_files_exsists_in_current_csi_version (){
```
oriyarde, line 26: exist*

hack/project_info.sh
```diff
@@ -0,0 +1,41 @@
...
+  current_csi_version=$(get_current_csi_version)
+  if ! compgen -G "${PWD}/deploy/olm-catalog/*/$current_csi_version" > /dev/null; then
+    exit 0
+  fi
+}
+
+get_csv_files (){
+  current_csi_version=$(get_current_csi_version)
+  ls deploy/olm-catalog/*/$current_csi_version/manifests/ibm-block-csi-operator.v$current_csi_version.clusterserviceversion.yaml
```
oriyarde, line 35: please [avoid wildcards](https://github.com/IBM/ibm-block-csi-operator/pull/220#discussion_r744253629)

oriyarde, line 35: I don't think this line would work with certified.
please update from develop and ensure certified is running and passing as well

hack/check-files-alignment.sh
oriyarde: I meant that we are missing a `verify_yaml_doc_in_multidoc_yaml` function that receives parameters

ArbelNathan: @oriyarde I agree that a more generic method is generally better.

ArbelNathan (dismissed):
LGTM

hack/check-files-alignment.sh
```diff
@@ -0,0 +1,78 @@
...
+}
+
+verify_yaml_doc_in_multidoc_yaml(){
+  yamls_dictionary=${1#*=}
+  eval "declare -A yaml_kinds_by_origin_yaml_path="${yamls_dictionary}
```
oriyarde, line 38: https://stackoverflow.com/a/3467959
>Avoid eval like the plague

hack/check-files-alignment.sh
```diff
@@ -0,0 +1,78 @@
...
+}
+
+verify_full_operator_yaml_is_aligned(){
+  echo "check full operator yaml alignment"
+  declare -A yaml_kinds_by_origin_yaml_path=(
+      [$roles_yaml_path]="ClusterRole"
+      [$crd_yaml_path]="CustomResourceDefinition"
+      ["config/rbac/service_account.yaml"]="ServiceAccount"
+      ["config/rbac/role_binding.yaml"]="ClusterRoleBinding"
+      ["config/manager/manager.yaml"]="Deployment"
```
oriyarde, line 52: why do we need the kinds hardcoded? can't we find them dynamically?

hack/check-files-alignment.sh
```diff
@@ -0,0 +1,78 @@
...
+  are_manifest_files_exist_in_current_csi_version
+  csv_files=$(get_csv_files)
+  for csv_file in $csv_files; do
+    diff <(yq e .rules $roles_yaml_path) <(yq e .spec.install.spec.clusterPermissions[0].rules $csv_file)
+  done
+}
+
+verify_no_crds_diff (){
+  echo "check crds alignment"
+  are_manifest_files_exist_in_current_csi_version
```
oriyarde, line 68: code dup

hack/project_info.sh
```diff
@@ -0,0 +1,42 @@
...
+
+are_manifest_files_exist_in_current_csi_version (){
+  if ! compgen -G "${PWD}/deploy/olm-catalog/*/$current_csi_version" > /dev/null; then
+    exit 0
+  fi
+}
+
+get_csv_files (){
+  ls deploy/olm-catalog/ibm-block-csi-operator-community/$current_csi_version/manifests/ibm-block-csi-operator.v$current_csi_version.clusterserviceversion.yaml
+  ls deploy/olm-catalog/ibm-block-csi-operator/$current_csi_version/manifests/ibm-block-csi-operator.clusterserviceversion.yaml
```
oriyarde, line 36: consider moving the olm-catalog path do a common "const"

hack/check-files-alignment.sh
matancarmeli7: I tried without it and I didn't manage to find a solution, the variable pass between the functions interrupted it, you sent an example without a function

hack/check-files-alignment.sh
matancarmeli7: I will open a ticket

hack/check-files-alignment.sh
```diff
@@ -0,0 +1,78 @@
...
+  project_dirname=ibm-block-csi-operator
+  cd ..
+  cp -r $project_dirname ./$project_dirname-expected
+  cd $project_dirname-expected/
+  make update
+  cd ..
+  diff -qr --exclude=bin $project_dirname $project_dirname-expected/
+  rm -rf $project_dirname-expected/
+  cd $project_dirname
+}
```
oriyarde, line 34:
```suggestion
check_generation ()(
  echo "check generation"
  project_path=$PWD
  cp -r $project_path $project_path-expected
  cd $project_path-expected
  make update
  diff -qr --exclude=bin $project_path $project_path-expected/
  cd ..
  rm -rf $project_path-expected/
)
```
would this work? less `cd` (+ absolute paths are more accurate)

hack/check-files-alignment.sh
oriyarde: to add `resource_type=$(yq e .kind $original_yaml)`?

hack/project_info.sh
oriyarde: I [still see the wildcard](https://github.com/IBM/ibm-block-csi-operator/blob/5f6371694ab49490297168fe2b3765156072b9dc/hack/project_info.sh#L29).
what happens if there's only one bundle for the current csi version? I don't think the test should fail. it should test what it can, and allow separate additions

hack/check-files-alignment.sh
oriyarde: I [don't think we need a dictionary](https://github.com/IBM/ibm-block-csi-operator/pull/226#discussion_r774186988)

PR #238, opened 2021-12-25
CSI-3671 add csv yamls

oriyarde:


deploy/olm-catalog/ibm-block-csi-operator-community/1.9.0/manifests/ibm-block-csi-operator.v1.8.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,549 @@
+apiVersion: operators.coreos.com/v1alpha1
```
ArbelNathan: v1.8.0 in the name.
and why in certified it has no version? could we align the naming formats?

oriyarde: I think that is the reason I missed it here in the first place.
we had to remove the version in 1.8.0 certified publication, due to red hat assertions

ArbelNathan (approved):
LGTM

PR #290, opened 2022-12-08
add pvc controller

matancarmeli7:
Signed-off-by: matancarmeli7 <matan.carmeli7@gmail.com>


ELENAGER (approved):
LGTM

volume-group-operator/controllers/persistentvolumeclaim/settings.go
```diff
@@ -0,0 +1,40 @@
...
+package persistentvolumeclaim
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+var (
+	persistentVolumeClaim = "PersistentVolumeClaim"
+	pvcPredicate          = predicate.Funcs{
```
ArbelNathan, line 26: what is this for?

matancarmeli7: that the controller will only watch update and create events, we don't need a delete event, because if a user wants to delete a PVC that is already in a `volumeGroup` he needs to remove it from the `volumeGroup`, which is an update event

ArbelNathan (approved):
LGTM

PR #294, opened 2022-12-11
CSI 4969 volumegroup controller implementation

ArbelNathan:


volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,163 @@ type VolumeGroupReconciler struct {
...
+	secret := make(map[string]string)
+	if secretName != "" && secretNamespace != "" {
+		secret, err = r.Utils.GetSecret(logger, secretName, secretNamespace)
+		if err != nil {
+			uErr := r.updateVolumeGroupStatusError(instance, logger, err.Error())
+			if uErr != nil {
+				logger.Error(uErr, "failed to update volumeGroup status", "VGName", instance.Name)
+			}
+
+			return reconcile.Result{}, err
```
ELENAGER, line 105: When request object not found - return nil and don't requeue. Like here:
https://github.com/csi-addons/volume-replication-operator/blob/main/controllers/volumereplication_controller.go#L82

volume-group-operator/controllers/utils/volumegroupcontent.go
```diff
@@ -64,6 +59,15 @@ func (r *ControllerUtils) generateVolumeGroupContentStatus(groupCreationTime *me
 	}
 }
 
+func (r *ControllerUtils) UpdateVolumeGroupStatus(logger logr.Logger, vgc *volumegroupv1.VolumeGroupContent, groupCreationTime *metav1.Time, ready *bool) error {
+	vgc.Status = r.generateVolumeGroupContentStatus(groupCreationTime, ready)
```
matancarmeli7, line 63: I think we should pass the `pvList`, I think with this code it will overwrite the `pvList` and leave it empty

volume-group-operator/controllers/utils/volumegroupcontent.go
```diff
@@ -64,6 +59,15 @@ func (r *ControllerUtils) generateVolumeGroupContentStatus(groupCreationTime *me
 	}
 }
 
+func (r *ControllerUtils) UpdateVolumeGroupStatus(logger logr.Logger, vgc *volumegroupv1.VolumeGroupContent, groupCreationTime *metav1.Time, ready *bool) error {
```
matancarmeli7, line 62: can you move `UpdateVolumeGroupStatus` function above `generateVolumeGroupContentStatus` function?

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+	if r.DriverConfig.DriverName != vgcObj.Driver {
+		return ctrl.Result{}, nil
+	}
+
+	err = utils.ValidatePrefixedParameters(vgcObj.Parameters)
+	if err != nil {
+		logger.Error(err, "failed to validate parameters of volumegroupClass", "VGClassName", vgcObj.Name)
+		uErr := r.updateVolumeGroupStatusError(instance, logger, err.Error())
+		if uErr != nil {
+			logger.Error(uErr, "failed to update volumeGroup status", "VGName", instance.Name)
```
matancarmeli7, line 97: can we move this log to `updateVolumeGroupStatus` function?, because it looks like all the functions that use `updateVolumeGroupStatus` are using this log on an error

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+		}
+
+		return ctrl.Result{}, err
+	}
+	parameters := utils.FilterPrefixedParameters(utils.VolumeGroupParameterPrefix, vgcObj.Parameters)
+
+	secretName := vgcObj.Parameters[utils.PrefixedVolumeGroupSecretNameKey]
+	secretNamespace := vgcObj.Parameters[utils.PrefixedVolumeGroupSecretNamespaceKey]
```
matancarmeli7, line 105: what will happen when the parameters in the volumeGroupClass won't have those keys? it will fail?

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+		secret, err = r.Utils.GetSecret(logger, secretName, secretNamespace)
+		if err != nil {
+			uErr := r.updateVolumeGroupStatusError(instance, logger, err.Error())
+			if uErr != nil {
+				logger.Error(uErr, "failed to update volumeGroup status", "VGName", instance.Name)
+			}
+
+			return reconcile.Result{}, err
+		}
+	}
```
matancarmeli7, line 117: can we move all this section to a `getSecret` file in the controller file that will return the secret and an error and just handle the error in the reconcile?

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+
+	if instance.GetDeletionTimestamp().IsZero() {
+		if err = r.Utils.AddFinalizerToVG(logger, instance); err != nil {
+			logger.Error(err, "Failed to add VolumeGroup finalizer")
```
matancarmeli7, line 121: IMO this can be in the `AddFinalizerToVG` function

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+				}
+			}
+
+			if err = r.Utils.RemoveFinalizerFromVG(logger, instance); err != nil {
+				logger.Error(err, "Failed to remove volume group finalizer")
+
+				return reconcile.Result{}, err
+			}
+
+		}
```
matancarmeli7, line 160: can you move this code to a function?

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+					logger.Error(err, "Failed to remove volume group content finalizer")
+
+					return reconcile.Result{}, err
+				}
+				if err = r.Client.Delete(context.TODO(), volumeGroupContent); err != nil {
+					logger.Error(err, "Failed to remove volume group content finalizer")
+
+					return reconcile.Result{}, err
+				}
+			}
```
matancarmeli7, line 152: and can you move this code to another function?

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+
+			return reconcile.Result{}, err
+		}
+
+	} else {
+		if utils.Contains(instance.GetFinalizers(), utils.VolumeGroupFinalizer) {
+			volumeGroupContent, uErr := r.Utils.GetVolumeGroupContent(logger, instance)
+			if uErr != nil {
+				if !errors.IsNotFound(uErr) {
+					logger.Error(uErr, "failed to get volumeGroupContent", *instance.Spec.Source.VolumeGroupContentName)
```
matancarmeli7, line 131: IMO this log should be in `GetVolumeGroupContent` function, (there are other logs in the reconcile that can be handled in their functions and shouldn't be handled in the reconcile)

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -70,3 +248,69 @@ func (r *VolumeGroupReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.D
...
+
+	if resp.Error != nil {
+		logger.Error(resp.Error, "failed to delete volume group")
+		return resp.Error
+	}
+
+	return nil
+}
+
+func (r *VolumeGroupReconciler) createVolumeGroup(volumeGroupName string, parameters, secrets map[string]string) *volumegroup.Response {
```
matancarmeli7, line 295:
```suggestion
func (r *VolumeGroupReconciler) createVolumeGroupResponse(volumeGroupName string, parameters, secrets map[string]string) *volumegroup.Response {
```

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+	vgc := r.Utils.GenerateVolumeGroupContent(volumeGroupName, instance, vgcObj, createVolumeGroupResponse, secretName, secretNamespace)
+
+	if err = r.Utils.CreateVolumeGroupContent(logger, instance, vgc); err != nil {
+		logger.Error(err, "failed to create volumeGroupContent", "VGCName", vgc.Name)
+		return reconcile.Result{}, err
+	}
+	instance.Spec.Source = volumegroupv1.VolumeGroupSource{
+		VolumeGroupContentName: &vgc.Name,
+		Selector:               r.volumeGroupLabelSelector(instance),
+	}
```
matancarmeli7, line 193: can you move to a function?

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
+		logger.Error(err, "failed to update instance")
+		return reconcile.Result{}, err
+	}
+	ready := true
+	instance.Status = volumegroupv1.VolumeGroupStatus{
+		BoundVolumeGroupContentName: &vgc.Name,
+		GroupCreationTime:           groupCreationTime,
+		Ready:                       &ready,
+		Error:                       nil,
+	}
```
matancarmeli7, line 205: can you move it to a function?

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,184 @@ type VolumeGroupReconciler struct {
...
 
-// SetupWithManager sets up the controller with the Manager.
+func makeVolumeGroupName(prefix string, volumeGroupUID string) (string, error) {
+	if len(volumeGroupUID) == 0 {
+		return "", fmt.Errorf("Corrupted volumeGroup object, it is missing UID")
+	}
+	return fmt.Sprintf("%s-%s", prefix, volumeGroupUID), nil
+}
+
+func (r *VolumeGroupReconciler) volumeGroupLabelSelector(instance *volumegroupv1.VolumeGroup) *metav1.LabelSelector {
```
matancarmeli7, line 236:
```suggestion
func (r *VolumeGroupReconciler) getVolumeGroupLabelSelector(instance *volumegroupv1.VolumeGroup) *metav1.LabelSelector {
```

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: no.. its only check for params with those keys from the first place

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: the function creates the vg on storage, not creating the response.

volume-group-operator/controllers/utils/volumegroupcontent.go
ArbelNathan: done

volume-group-operator/controllers/utils/volumegroupcontent.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +61,163 @@ type VolumeGroupReconciler struct {
...
+		if utils.Contains(instance.GetFinalizers(), utils.VolumeGroupFinalizer) {
+			volumeGroupContent, uErr := r.Utils.GetVolumeGroupContent(logger, instance)
+			if uErr == nil {
+				volumeGroupId := volumeGroupContent.Spec.Source.VolumeGroupHandle
+				if err = r.deleteVolumeGroup(logger, volumeGroupId, secret); err != nil {
+					logger.Error(err, "failed to delete volume group")
+
+					return ctrl.Result{}, err
+				}
+			}
```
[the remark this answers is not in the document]
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
matancarmeli7: oh sorry right, my bad

volume-group-operator/controllers/volumegroup_controller.go
matancarmeli7: where do we validate that it checks for params with those keys before?

ELENAGER (approved):
LGTM

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -70,3 +211,47 @@ func (r *VolumeGroupReconciler) SetupWithManager(mgr ctrl.Manager, cfg *config.D
...
 }
+
+func (r *VolumeGroupReconciler) deleteVolumeGroup(logger logr.Logger, volumeGroupId string, secrets map[string]string) error {
+	c := volumegroup.CommonRequestParameters{
+		VolumeGroupID: volumeGroupId,
+		Secrets:       secrets,
+		VolumeGroup:   r.VolumeGroupClient,
+	}
+
+	volumeGroup := volumegroup.VolumeGroup{
```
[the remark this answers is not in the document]
matancarmeli7: so it will look like this:  `volumegroup.New()`

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -18,24 +18,26 @@ package controllers
```
matancarmeli7, line 18: where do we handle the events we should add to the `volumeGroup` objects in Kubernetes, both in success and failure?

volume-group-operator/controllers/utils/volumegroup.go
```diff
@@ -0,0 +1,47 @@
+package utils
+
+import (
+	"github.com/IBM/volume-group-operator/api/v1"
+	volumegroupv1 "github.com/IBM/volume-group-operator/api/v1"
```
matancarmeli7, line 5: can you remove one of them?

matancarmeli7, line 5: and we need to decide how do we want to call our API from the code (can be in a ticket), because we call it [`volumegroupv1`](https://github.com/IBM/ibm-block-csi-operator/blob/feature/CSI-4969_volumegroup_controller_implementation/volume-group-operator/controllers/utils/volumegroup.go#L5) and [`csiv1`](https://github.com/IBM/ibm-block-csi-operator/blob/feature/CSI-4969_volumegroup_controller_implementation/volume-group-operator/main.go#L38)

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +55,153 @@ type VolumeGroupReconciler struct {
...
+	secretNamespace := vgcObj.Parameters[utils.PrefixedVolumeGroupSecretNamespaceKey]
+	secret := make(map[string]string)
+	if secretName != "" && secretNamespace != "" {
+		secret, err = utils.GetSecretData(r.Client, logger, secretName, secretNamespace)
+		if err != nil {
+			uErr := utils.UpdateVolumeGroupStatusError(r.Client, instance, logger, err.Error())
+			if uErr != nil {
+				return ctrl.Result{}, uErr
+			}
+			return reconcile.Result{}, err
```
matancarmeli7, line 106: what is the difference between `ctrl.Result{}` to `reconcile.Result{}`, I think we should choose one

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +55,153 @@ type VolumeGroupReconciler struct {
...
+
+	groupCreationTime := getCurrentTime()
+	volumeGroupName, err := makeVolumeGroupName(utils.VolumeGroupPrefix, string(instance.UID))
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	createVolumeGroupResponse := r.createVolumeGroup(volumeGroupName, parameters, secret)
+	if createVolumeGroupResponse.Error != nil {
+		logger.Error(err, "failed to create volume group")
```
matancarmeli7, line 134:
```suggestion
		logger.Error(createVolumeGroupResponse.Error, "failed to create volume group")
```

volume-group-operator/controllers/volumegroup_controller.go
```diff
@@ -53,14 +55,153 @@ type VolumeGroupReconciler struct {
...
+	if err = utils.CreateVolumeGroupContent(r.Client, logger, instance, vgc); err != nil {
+		return reconcile.Result{}, err
+	}
+	utils.UpdateVolumeGroupSource(instance, vgc)
+
+	if err = utils.UpdateObject(r.Client, instance); err != nil {
+		return reconcile.Result{}, err
+	}
+	uErr := utils.UpdateVolumeGroupStatus(r.Client, instance, vgc, groupCreationTime, true, logger)
+	if uErr != nil {
```
matancarmeli7, line 153: can you handle this error?

volume-group-operator/controllers/volumegroup_controller.go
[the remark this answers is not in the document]
ArbelNathan: It make no sence to me, volumegroup contains several types that are part of it and also public and used outside of it.

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done, tnx

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/utils/volumegroup.go
ArbelNathan: done

ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: before what? I don't understand the question, those are special prefixes to the sidecar, we are checking that they are not empty.. every other parameter is optional and we don't care about it.
it is not suggested that the secret is mandatory, just that the parameter that we managing is valid.

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: in the `Error` field in `Status`

matancarmeli7: this is not event, if there is a newer Error the user won't see the previous ones

matancarmeli7: there is a difference between field in the Status than Event of an object

volume-group-operator/controllers/volumegroup_controller.go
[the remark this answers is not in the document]
matancarmeli7: so why did you do it [here](https://github.com/IBM/ibm-block-csi-operator/blob/feature/CSI-4969_volumegroup_controller_implementation/volume-group-operator/pkg/client/volumegroup_client.go#L38)? it's the same use case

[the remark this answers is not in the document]
ArbelNathan: is bit more literal than just `New`

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: the user won't see any event in the long run, they will all disappear. the Error in Status will remain till resolved.
anyway, If you find a way to add Events easily fill free to add them, I didn't.

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: done

volume-group-operator/controllers/utils/volumegroup.go
```diff
@@ -0,0 +1,46 @@
...
+		logger.Error(err, "failed to update status")
+
+		return err
+	}
+	return nil
+}
+
+func UpdateVolumeGroupStatus(client client.Client, instance *volumegroupv1.VolumeGroup, vgc *volumegroupv1.VolumeGroupContent,
+	groupCreationTime *metav1.Time, ready bool, logger logr.Logger) error {
+	instance.Status = volumegroupv1.VolumeGroupStatus{
```
matancarmeli7, line 28: can you add all the Status fields? because it overwrites fields that do not appear here, can you also do it for the `volumeGroupContent` and class if needed?

volume-group-operator/controllers/volumegroup_controller.go
matancarmeli7: it's part of the spec and part of your ticket, you don't create objects and don't create events for them, it's a blocker for this PR IMO

volume-group-operator/controllers/volumegroup_controller.go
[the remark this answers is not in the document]
matancarmeli7: the `New` is enough because you write before it the type of the object (`volumegroup`), but also function with another name is good

volume-group-operator/controllers/volumegroup_controller.go
matancarmeli7: [this](https://pkg.go.dev/k8s.io/client-go/tools/events#EventSinkImpl.Create) doesn't work?

matancarmeli7: I managed to create events [here](https://github.com/IBM/ibm-block-csi-operator/blob/feature/CSI-4972_pvc_controller_remove_pvc_from_vg_with_grpc/volume-group-operator/controllers/utils/event.go) you can use this code

ArbelNathan: please link the spec requirement for the events.

ArbelNathan: I opened a ticket for it CSI-5149

matancarmeli7: ticket is not good, we need to do it in this version, and it's known thing for every Kubernetes object, you can't create one without events

matancarmeli7: if the ticket will be in 1.11 version so it's fine, but if not, we can't GA that way

volume-group-operator/controllers/utils/volumegroup.go
ArbelNathan: done

volume-group-operator/controllers/volumegroup_controller.go
ArbelNathan: it is, but anyway I disagree that it's blocking. its part of no spec and replication sidecar doesn't even have more events than the Error in status

volume-group-operator/controllers/volumegroup_controller.go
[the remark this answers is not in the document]
ArbelNathan: done

volume-group-operator/controllers/utils/volumegroup.go
matancarmeli7: what about the PVCList and PVList fields in the `volumeGrouo` and `volumeGroupContent`?

ArbelNathan: I changed it so no we did not override the Status but only change the fields we need on the existing one

volume-group-operator/controllers/volumegroup_controller.go
matancarmeli7: if the replication sidecar doesn't do it it doesn't mean it is good, we need to be better than them, and I don't think we can GA this feature without events so we must do this ticket and it's not that big of a change because I did most of the job already

ELENAGER: Arbel opened ticket for adding events and this ticket is important enough to handle it until GA. We still have 2 sprints to do it. I don't think this discussion is productive, let's merge this PR and then Arbel will be able to proceed with events ticket.

matancarmeli7: feature freeze is in this sprint, so no new content after this sprint

matancarmeli7: so we don't have two sprints for it

volume-group-operator/controllers/volumegroup_controller.go
matancarmeli7: never mind my bad

matancarmeli7 (approved):
LGTM GJ, IMO we need to address the event ticket asap

volume-group-operator/controllers/volumegroup_controller.go
ELENAGER: feature freeze is in this sprint can't happen, unless we decided not to come with replication for volume groups

matancarmeli7: currently, the feature freeze is in this sprint and we should act like this until Eran will change it

PR #323, opened 2023-03-07
CSI-5299_change provisioner version

samishekh:
change csi-provisioner version in CR to match version in csv


matancarmeli7 (approved):
LGTM

PR #342, opened 2023-06-06
Feature/csi 5610 add an option to enable/disable call home

matancarmeli7:


ArbelNathan (approved):
LGTM

leonid-s-usov:
Wait, so what happens to existing installations? Will they automatically receive the new spec field set to true? I wonder how it works with operator upgrades.

PR #455, opened 2025-11-21
CSI-5938:Host Definer upgrade doesn't use the correct image

sumitguptaibm:
Hostdefiner image repo is: [quay.io/ibmcsiblock/ibm-block-csi-host-definer](http://quay.io/ibmcsiblock/ibm-block-csi-host-definer) , 
But hostdefiner_controller was not treating this image as official image. Since the image was getting treated as unofficial, CSI-Operator was not upgrading Hostdefiner. 
If we update Hostdefiner_controller logic to treat  [quay.io/ibmcsiblock/ibm-block-csi-host-definer](http://quay.io/ibmcsiblock/ibm-block-csi-host-definer) as official image using below switch case, operator would start upgrading Hostdefiner component to the desired version along with CSI-Driver and CSI-Node.


shlomitn:
The change is already in 1.13.0.

